PR #1, opened 2018-12-13
Add docs for API.

lachmanfrantisek:


ogr/services/abstract.py
```diff
@@ -4,53 +4,161 @@ def __init__(self):
...
+        :param remote_url: str
+        :return: GitService
+        """
         raise NotImplementedError()
 
     def get_project(self, namespace=None, user=None, repo=None):
+        """
+        Get the GitProject instance
+
+        :param namespace: str
```
TomasTomecek: aren't namespace and user the same thing? is this just to make it more clear?

ogr/services/abstract.py
```diff
@@ -4,53 +4,161 @@ def __init__(self):
...
+
+        :param title: str
+        :param body: str
+        :param target_branch: str
+        :param current_branch: str
+        :return: id of the new pull-request
+        """
+        raise NotImplementedError()
+
+    def get_pr_list(self, status="open"):
```
TomasTomecek: it would be awesome if status was enum

ogr/services/abstract.py
```diff
@@ -4,53 +4,161 @@ def __init__(self):
...
+            "author":"???",
+            "created": "???"
+            "edited": "???"
+        }]
+        """
         raise NotImplementedError()
 
     def pr_comment(self, pr_id, body, commit=None, filename=None, row=None):
+        """
+        Add new commit to the pull request.
```
TomasTomecek: comment*

PR #2, opened 2019-01-10
Update API and Pagure

lachmanfrantisek:
- Add tests for Pagure and OurPagure.
- Update the API and Pagure implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,6 @@
 
 Library for one API for many git forges. (e.g. github, gitlab, pagure).
 
+# Contribution notes
+
+- Property should not connect to network.
--- a/ogr/services/abstract.py
+++ b/ogr/services/abstract.py
@@ -40,6 +40,42 @@ class GitService:
 
 
 class GitProject:
+
+    def __init__(self, repo, namespace, service):
+        """
+        :param repo: name of the project
+        :param namespace:   github: username or org name
+                            gitlab: username or org name
+                            pagure: namespace (e.g. "rpms")
+                                for forks: "fork/{username}/{namespace}"
+        :param service: GitService instance
+        """
+        self.service = service
+        self.repo = repo
+        self.namespace = namespace
+
+    def is_forked(self):
+        """
+        True, if the project is forked by the user.
+
+        :return: Bool
+        """
+        raise NotImplementedError()
+
+    @property
+    def is_fork(self):
+        """True if the project is a fork."""
+        raise NotImplementedError()
+
+    @property
+    def full_repo_name(self):
+        """
+        Get repo name with namespace
+
+        :return: str
+        """
+        return f"{self.namespace}/{self.repo}"
+
     def get_branches(self):
         """
         List of project branches.
@@ -56,15 +92,11 @@ class GitProject:
         """
         raise NotImplementedError()
 
-    def pr_create(self, title, body, target_branch, current_branch):
+    def get_fork(self):
         """
-        Create a new pull request.
+        GitProject instance of the fork if the fork exists, else None
 
-        :param title: str
-        :param body: str
-        :param target_branch: str
-        :param current_branch: str
-        :return: PullRequest
+        :return: GitProject or None
         """
         raise NotImplementedError()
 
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -2,3 +2,6 @@
 
 Library for one API for many git forges. (e.g. github, gitlab, pagure).
 
+# Contribution notes
```
TomasTomecek: This section would use some more love but that's probably out of scope of this PR.

ogr/services/abstract.py
```diff
@@ -40,6 +40,42 @@ def user(self):
...
+
+    @property
+    def is_fork(self):
+        """True if the project is a fork."""
+        raise NotImplementedError()
+
+    @property
+    def full_repo_name(self):
+        """
+        Get repo name with namespace
```
TomasTomecek: please add an example so everyone know what exactly this should yield

ogr/services/abstract.py
```diff
@@ -95,6 +127,18 @@ def get_pr_comments(self, pr_id):
         """
         raise NotImplementedError()
 
+    def pr_create(self, title, body, target_branch, current_branch):
```
TomasTomecek: how about target remote? I may suggest to rename current_branch to source_branch

ogr/services/abstract.py
```diff
@@ -56,15 +92,11 @@ def get_description(self):
         """
         raise NotImplementedError()
 
-    def pr_create(self, title, body, target_branch, current_branch):
+    def get_fork(self):
```
TomasTomecek: Even though you say that properties should not access internet, I'd suggest to cache this since forks likely not change for a user.

ogr/services/github.py
```diff
@@ -43,8 +43,8 @@ def create_from_remote_url(cls, remote_url, **kwargs):
     @staticmethod
     def is_fork_of(user_repo, target_repo):
         """ is provided repo fork of gh.com/{parent_repo}/? """
-        return user_repo.fork_create and user_repo.parent and \
-               user_repo.parent.full_name == target_repo
+        return user_repo.fork_create and user_repo.get_parent and \
```
TomasTomecek: isn't get_parent meant to be a method?

ogr/services/pagure.py
```diff
@@ -140,41 +151,72 @@ def pr_create(self, title, body, target_branch, source_branch):
...
-    @property
     def is_forked(self):
-        return self.fork is not None
+        pass
 
     @property
-    def git_urls(self):
-        return self.pagure.git_urls
+    def is_fork(self):
+        return "fork" in self.namespace
```
TomasTomecek: cheezy

ogr/services/abstract.py
lachmanfrantisek: Do you mean caching the result forever? Or some "refresh=False" argument?

Do we want a property or method?

ogr/services/github.py
lachmanfrantisek: Thanks, good catch. Boolean methods did not rise error.

ogr/services/pagure.py
lachmanfrantisek: I know. The fork/namespace/user logic will probably change when we start implementing other services.

ogr/services/abstract.py
TomasTomecek: good questions; I would probably only cache a real value, for None, I would keep retrying

let's leave it as it is for now and we can improve that in future

ogr/services/pagure.py
TomasTomecek: no worries, keep it as it is; each implementation can override it easily

ogr/services/abstract.py
lachmanfrantisek: > target remote

Since we are working with `upstream repo` <-> `fork` relation, we do not need to know it.

- When we have PR on the upstream project, the target remote is the repo itself.
- When we have forked project, we are creating PR against parent project.

I'm not sure about the multiple-remote situations. It can be hard to make it compatible between different services.

lachmanfrantisek:
I've pushed the changes. Is there anything else that needs to be done?

ogr/services/abstract.py
TomasTomecek: I didn't realize, sounds good then!

lachmanfrantisek:
Thanks! I'm going to push another with setup.py...

PR #3, opened 2019-01-10
Introduce packaging

lachmanfrantisek:
- add MIT license
- use setuptools-scm
- add Makefile target for image and pypi-checking

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Red Hat
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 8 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Userspace Containerization team at Red Hat
```
TomasTomecek: just red hat, we are not a separate enitity :D

Makefile
```diff
@@ -0,0 +1,25 @@
...
+	sudo ansible-bender build --build-volumes $(CURDIR):/src:Z -- ./recipe.yaml $(BASE_IMAGE) $(OGR_IMAGE)
+
+check:
+	PYTHONPATH=$(CURDIR) pytest-3 -v $(TEST_TARGET)
+
+shell:
+	sudo podman run --rm -ti -v $(CURDIR):/src:Z -w /src $(OGR_IMAGE) bash
+
+check-pypi-packaging:
+	sudo podman run --rm -ti -v $(CURDIR):/src:Z -w /src $(OGR_IMAGE) bash -c '\
```
TomasTomecek: is base_image meant to be used here? won't we run into the same issue you described to me yesterday?

recipe.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - python3-setuptools_scm
+      - python3-setuptools_scm_git_archive
+      - python3-wheel  # for bdist_wheel
+  - name: Install latest twine for sake of check command
+    pip:
+      name:
+      - twine  # we need newest twine, b/c of the check command
+      - readme_renderer[md]
+      state: latest
+  - name: Install dependencies for CI watcher
```
TomasTomecek: we won't need this task

Makefile
TomasTomecek: just tried this and it works

LICENSE
lachmanfrantisek: I've copied that from source-git -- we need to change that also there.

lachmanfrantisek: It is probably in multiple repos...

Makefile
lachmanfrantisek: The problem was due to these lines:

```

setuptools.setup(use_scm_version=True,
                 packages=setuptools.find_packages(exclude=['tests*']), )
```

I used `exclude=['tests']`, but that was wrong, beacuse it returns e.g. `tests.unit` and `tests.integration`...

lachmanfrantisek: This needs fix on the sourcegit side as well.

recipe.yaml
lachmanfrantisek: Oh, I missed that. Thanks.

lachmanfrantisek: But we need most of the packages, so I've changed that only a bit.

lachmanfrantisek:
I've fixed the issues from the review.

LICENSE
TomasTomecek: yup

PR #7, opened 2019-01-29
Add type annotations and fix create_pr

lachmanfrantisek:
- Remove duplicated test.
- Add annotations.


lachmanfrantisek:
Oh, that's nothing with this PR. The test files do not start with `test`. I'll correct that.

lachmanfrantisek:
> That has a lot to do with this PR :D I can't verify without tests.

Just mean, that it does not work also before..;-) I am working on it and have problems with cyclic dependencies due to the typing.

Do you know, how to type this? It needs to imported GitService before importing GitService..;-(

```
class GitService:

    @classmethod
    def create_from_remote_url(cls, remote_url) -> GitService:
        raise NotImplementedError()
```

lachmanfrantisek:
I've fixed the following:
- rename test files to be found by pytest
- split abstract classes to `ogr.abstract` and `ogr.services.base`
    - abstract does not contain any real code
    - base classes contains common code for all services
- use `from __future__ import annotations` to fix cyclic problems

lachmanfrantisek:
Thanks! The tests need some care.

PR #30, opened 2019-03-04
[integration tests] skip whole module if not all env. variables set

jscotka:


tests/integration/test_github.py
```diff
@@ -6,6 +6,17 @@
 from ogr.abstract import PRStatus
 from ogr.services.github import GithubService
 
+def requirements_met():
```
TomasTomecek: if you add this in conftest.py, pytest will inject it into every test module, no need to have it three times

jscotka: @TomasTomecek that sounds interesting, I've did not know that it is possible.

But there are and could be various requirements, eg. test_github needs to have set ``GITHUB_TOKEN`` and ``GITHUB_USER``, and pagure tests has ``PAGURE_USER`` and token, and eg. for upstream pagure you theoretically need to have 3 tokens, although for dowstream you need just one. so that is reason why this is probably better solution as I think.

tests/integration/test_github.py
```diff
@@ -6,6 +6,16 @@
 from ogr.abstract import PRStatus
 from ogr.services.github import GithubService
 
+def skip_tests(env_vars_list):
+    requirements_met = True
+    for item in env_vars_list:
+        if os.environ.get(item) is None:
+            requirements_met = False
```
jpopelka:
```python
  requirements_met = all(os.environ.get(item) for item in env_vars_list)
```

tests/integration/test_github.py
```diff
@@ -6,6 +6,16 @@
...
+    return pytest.mark.skipif(not requirements_met,
+                              reason="you have to have set env vars: {}".format(env_vars_list))
+
+pytestmark = skip_tests(["GITHUB_TOKEN", "GITHUB_USER"])
```
jpopelka: How does this `pytestmark` work? Is it documented somewhere? I can't find anything.

tests/integration/test_github.py
jpopelka: `conftest.py` is IMHO used for fixtures only, pytest does not import functions from `conftest.py` into other modules - but I might be wrong.
But we can put `skip_tests()` into `conftest.py` and import it explicitly to other modules
```python
from .conftest import skip_tests
```

And in each module we would just specify the requirements
```python
pytestmark = skip_tests(["GITHUB_TOKEN", "GITHUB_USER"])
```

tests/integration/test_github.py
jscotka: yep, this could be used, it is not really same, same could be: replaced it with ``os.environ.get(item) is not None`` because ``A=""``  will be transformed to False, although it is defined.

TomasTomecek:
also, how the F did you push into Honza's branch? Are you a magician?

jpopelka:
@TomasTomecek read the line below the latest comment - slowly & carefully :-)

PR #34, opened 2019-03-15
pre-commit config & black/Flake8/mypy fixes

jpopelka:


ogr/services/github.py
```diff
@@ -209,6 +209,8 @@ def get_releases(self) -> List[Release]:
 
 
 class GithubUser(BaseGitUser):
+    service: GithubService
+
```
jpopelka: previously the class looked like:
```python
class GithubUser(BaseGitUser):
    def __init__(self, service: GithubService) -> None:
        super().__init__(service=service)

    def get_username(self) -> str:
        return self.service.github.get_user().login
```
and mypy kept telling me that:
`error: "GitService" has no attribute "github"`
while pointing at the
`return self.service.github.get_user().login`
line.

I resolved it by declaring this instance variable.
@lachmanfrantisek do you have any better idea?

ogr/abstract.py
```diff
@@ -320,7 +320,7 @@ def get_file_content(self, path: str, ref="master") -> str:
 
 
 class GitUser:
-    def __init__(self, service: "GitService") -> None:
+    def __init__(self, service: GitService) -> None:
```
lachmanfrantisek: I think this can cause import problems since GitService is in the same file.

ogr/services/github.py
lachmanfrantisek: Isn't this a class attribute or is this used only for the typing?

jpopelka: It's just hint/annotation, see https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations

ogr/abstract.py
jpopelka: I thought we need to use the quotations only
* in the same class
* in the same module if the type/class is not yet defined

In this case, the `GitService` is defined on line 78, while this is line 323.

I can change it back, but I'd like to understand it.

lachmanfrantisek: Yes, you're right. It's already defined so it's not needed.

PR #40, opened 2019-03-20
WIP: Simulation

jscotka:


ogr/services/mock/github_mock.py
```diff
@@ -0,0 +1,21 @@
+# mocking github interface
+import github
+from ogr.services.mock import core
+
+
+class MockGitHubProject(core.MockClass):
```
lachmanfrantisek: Let's make it more general:
```suggestion
class MockProject(core.MockClass):
```

ogr/services/mock/core.py
```diff
@@ -0,0 +1,82 @@
...
+                if args:
+                    output += ", "
+                output += "**{}".format(kwargs)
+        output += (")")
+        write_output(output)
+        return output
+
+
+class MockClass(object):
+    configuration_file = "mock.yaml"
```
lachmanfrantisek: I would prefer not to use any files at all.

(The main use case, for now, is `--dry-run` in packit -- print only the called methods.)

ogr/services/mock/github_mock.py
lachmanfrantisek: And mock the `AbstractService`/`AbstractProject`/`AbstractUser`...

ogr/services/mock/core.py
```diff
@@ -0,0 +1,82 @@
...
+import inspect
+import yaml
+
+def write_output(*args):
+    print(*args)
+
+class CallLogger(object):
+    def __init__(self, namespace, class_obj):
+        self.namespace = namespace
+        self.class_obj = class_obj
```
lachmanfrantisek: I am not sure if I got the purpose of this class.

If it is for easier logging, why we have `namespace` attribute here -- it's specific to the `GithubProject/`PagureProject` classes.

ogr/services/github.py
```diff
@@ -8,9 +8,10 @@
 
 from ogr.abstract import GitUser, GitProject, PullRequest, PRComment, PRStatus, Release
 from ogr.services.base import BaseGitService, BaseGitProject, BaseGitUser
-
+from ogr.services.mock.github_mock import MockGithub
 logger = logging.getLogger(__name__)
 
+github.Github = MockGithub
```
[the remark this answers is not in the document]
jscotka: Sure, this is there just for my testing, to not bother now how it will be switched.

ogr/services/mock/github_mock.py
jscotka: Yep, we can discuss then what everything we need to mock. Maybe we can strast with some minimal mock and them extend it.
 rename it to ``MockProject`` but this is mocking github project, and in case ``libpagure`` has also project, it could have another methods, so checking method names them will not work

lachmanfrantisek: I would like to have it as another implementation of the `AbstractProject`. All implementation (`GithubProject`/`PagureProject`) should implement the same methods -- this is the purpose of the OGR: One API for all git-forges.

ogr/services/mock/core.py
```diff
@@ -0,0 +1,82 @@
+import os
+import inspect
+import yaml
+
+def write_output(*args):
```
TomasTomecek: why not use print right away?

ogr/services/mock/core.py
```diff
@@ -0,0 +1,82 @@
...
+            if len(args) == 1:
+                output += str(args)[1:-2]
+            elif len(args) > 1:
+                output += str(args)[1:-1]
+            if kwargs:
+                if args:
+                    output += ", "
+                output += "**{}".format(kwargs)
+        output += (")")
+        write_output(output)
```
TomasTomecek: please use logger instead

ogr/services/mock/mock.yaml
```diff
@@ -0,0 +1,4 @@
+generic:
```
TomasTomecek: Why not have this in the code directly?

ogr/services/mock/core.py
jscotka: yes, this is optional. and main purpose is to store return values for testing, for dry-run it is not useful.

ogr/services/mock/core.py
jscotka: yes, the purpose is to see difference in log then as you mentioned `Github/Pagure`

ogr/services/mock/core.py
jscotka: Sure, I justed just wrap print command and easily replace it with some logger, or whatever

ogr/services/mock/core.py
jscotka: yep, it will be used there

ogr/services/mock/core.py
lachmanfrantisek: I would place testing stuff to `./tests/` and not mess with the regular code.

ogr/services/mock/mock.yaml
jscotka: My imagination is, to have there some templates, eg. as now there is integration testsuite stored `conu` and `colin` project, so that have there return for them.

And also I can imagine to use it also for storing values, in sense of some `--dryrun` with storing values on some internal place.

jscotka:
> The proposed implementation looks like a overkill to me. Let's discuss in person on the Mon architecture call.
> 
> As you said, you're basically reimplementing here existing mock libraries - I don't understand why such thing is needed. Just use mock or flexmock directly.

@TomasTomecek , I though about that  and `flexmock` seems little bit different and not so powerful.

`unittest.mock` library is more less very similar to my code, for this purpose with usage like `mock.Mock(spec=github.Github)` , but I had no experiences and never heard about that with that and code what I've wrote is not so complex and helped me  to understand what happens there also use these configuration yaml files to store predefined output.

But is it true that I could replace my `generic Mock ` class with this one and just extend it.  for my purpose, and avoid to create another mock class implementation, maybe not sure If I can do whatever I want. means output what class, module I'm mocking. I try to investigate it and maybe use it. Maybe I'll had to look on internal code of this, to see what I can do there. Or if you have experience I'll glad to hear about that and usages in these cases.

ogr/services/mock/mock.yaml
TomasTomecek: > My imagination is, to have there some templates, eg. as now there is integration testsuite stored `conu` and `colin` project, so that have there return for them.

I'm sorry but I don't understand.

> And also I can imagine to use it also for storing values, in sense of some `--dryrun` with storing values on some internal place.

You can't be serious: changing files in sitelib? `rpm -V` would not be happy and this could result into some pretty bad errors or race conditions. Runtime data are meant to be stored elsewhere - also I don't understand why should a library store data.

TomasTomecek:
I was actually expecting a PR similar to this one on release-bot: https://github.com/user-cont/release-bot/pull/177

Just adding a new flag dry_run and not change anything when dry_run is on.

ogr/services/mock/mock.yaml
jscotka: Ah sorry, this location is just for my purposes. It will not be here, I have it there just for debugging, then it could be stored beside tests, or whenever.

jscotka:
> I was actually expecting a PR similar to this one on release-bot: [user-cont/release-bot#177](https://github.com/user-cont/release-bot/pull/177)
> 
> Just adding a new flag dry_run and not change anything when dry_run is on.

@TomasTomecek yep, but this `dry-run` option is something little bit different, and `dry run` option will be implemented as part of `packit` not part of `org`, `ogr` just provide some mocking of env for another tools.

lachmanfrantisek:
> @TomasTomecek yep, but this dry-run option is something little bit different, and dry run option will be implemented as part of packit not part of org, ogr just provide some mocking of env for another tools.

Yes, the `--dry-run` would be part of packit, but we were speaking about "mock" implementation of the ogr API, that could be used from packit for this purpose.

TomasTomecek:
Okay, I leave this one to @lachmanfrantisek and @jpopelka then. My assumption was that ogr classes would accept dry_run argument which would not let them write to remote services. If we choose path of mocking all the methods, then we'll have to maintain twice as much code which will be such a pain. We would also need to be sure that methods are in sync.

lachmanfrantisek:
> Okay, I leave this one to @lachmanfrantisek and @jpopelka then. My assumption was that ogr classes would accept dry_run argument which would not let them write to remote services. If we choose path of mocking all the methods, then we'll have to maintain twice as much code which will be such a pain. We would also need to be sure that methods are in sync.

It would be nice if we can `mock`/`flexmock` the abstract classes automatically. I hope that some solution already exists.

jscotka:
This PR could be closed,
used simplier solution https://github.com/packit-service/ogr/pull/48
and provided also `read-only` mode there

PR #44, opened 2019-03-25
m0ar forking

### LINKED ISSUE #31: create_fork should not fail if the fork already exists
```
        if req.status_code != 200:
            LOG.error(output)
            if 'error_code' in output:
>               raise APIError(output['error'])
E               libpagure.exceptions.APIError: Repo "forks/ttomecek/sen" already exists     
```

```
/home/tt/.local/lib/python3.7/site-packages/ogr/services/pagure.py:151: in fork_create
    self._pagure.create_fork()
/home/tt/.local/lib/python3.7/site-packages/ogr/services/our_pagure.py:254: in create_fork                                                                                  
    data={"repo": self.repo_name, "namespace": self.namespace, "wait": True},
```

WDYT?

TomasTomecek:
* implement forking interface for github
* upgrade fork API

Fixes #31


ogr/services/github.py
```diff
@@ -66,8 +97,27 @@ def get_branches(self) -> List[str]:
     def get_description(self) -> str:
         return self.github_repo.description
 
-    def get_fork(self) -> Optional[GitProject]:
-        raise NotImplementedError
+    def get_fork(self, create: bool = True) -> Optional[GitProject]:
```
jpopelka:
```suggestion
    def get_fork(self, create: bool = True) -> Optional[GithubProject]:
```

tests/integration/test_github.py
```diff
@@ -1,6 +1,7 @@
 import os
 
 import pytest
+from github import GithubException
 from libpagure import APIError
```
jpopelka:
```suggestion
```

ogr/services/pagure.py
```diff
@@ -150,10 +151,7 @@ def pr_create(
     def fork_create(self) -> None:
         self._pagure.create_fork()
```
jpopelka: Shouldn't `fork_create()` return `PagureProject` ?

TomasTomecek:
> run the linters/checkers in CI

I swear I wanted to run it, but forgot. Okay, I install & try once more.

Seriously: I think we should run them as part of the CI process.

ogr/services/github.py
TomasTomecek: so I ran it locally to get used to it; and it's installed also!

ogr/abstract.py
```diff
@@ -132,11 +132,11 @@ def __init__(self, repo: str, service: GitService, namespace: str) -> None:
         self.repo = repo
         self.namespace = namespace
 
-    def is_forked(self) -> bool:
+    def is_forked(self) -> Optional["GitProject"]:
```
lachmanfrantisek: The method starting with `is` should return bool. This would be confusing.

ogr/services/github.py
```diff
@@ -66,8 +99,27 @@ def get_branches(self) -> List[str]:
...
+        if self.is_fork:
+            return None
+        f = self.is_forked()
+        if not f:
+            if create:
+                return self.fork_create()
+            else:
+                raise OgrException(
+                    f"Fork of {self.github_repo.full_name}"
+                    " does not exist and we were asked not to create it."
```
lachmanfrantisek: I would expect `None` in this case.

(In Gitlab context, the forking can be nested and "is_fork == not is_forked" is not always true.)

ogr/services/github.py
```diff
@@ -66,8 +99,27 @@ def get_branches(self) -> List[str]:
...
+
+        Returns None if this is a fork.
+
+        :param create: create a fork if it doesn't exist
+        :return: instance of GithubProject
+        """
+        if self.is_fork:
+            return None
+        f = self.is_forked()
+        if not f:
```
lachmanfrantisek: That's the reason I do not like one-letter variables and `is_something` returning an object...

ogr/abstract.py
TomasTomecek: Well you can always do `if git_project.is_forked()`...

My point here is that I want to get the fored_project as soon as possible: but the truth is that I can call get_fork() instead.

I'll change it.

ogr/services/github.py
TomasTomecek: ok

TomasTomecek: I guess you can also do forks of forks on github

ogr/services/github.py
TomasTomecek: that's why we're using typing, right? You can always check what you're getting

lachmanfrantisek: `Readability counts.` vs `Simple is better than complex.`..;-)

I just like the code that I can understand/review without any further observation or running the code.

TomasTomecek: I renamed

TomasTomecek:
@lachmanfrantisek should be updated now

TomasTomecek:
(let me rebase)

tests/integration/test_github.py
```diff
@@ -153,7 +153,7 @@ def test_commit_flags(colin_project):
 
 
 def test_fork(colin_project_fork):
-    assert colin_project_fork.is_fork
+    assert colin_project_fork.is_fork is True
```
jpopelka: oh, common

tests/integration/test_github.py
```diff
@@ -182,7 +182,7 @@ def test_create_fork(colin_project):
 
 def test_is_fork(colin_project):
     assert not colin_project.is_fork
-    assert colin_project.is_forked()
+    assert colin_project.is_forked() is True
```
jpopelka: seriously?

ogr/services/github.py
```diff
@@ -37,17 +41,50 @@ def change_token(self, new_token: str) -> None:
...
+        else:
+            self.github_repo = service.github.get_repo(
+                full_name_or_id=f"{namespace}/{repo}"
+            )
+
+    def _construct_fork_project(self) -> Optional["GithubProject"]:
+        gh_user = self.service.github.get_user()
+        user_login = gh_user.login
+        try:
+            return GithubProject(self.repo, self.service, namespace=gh_user.login)
```
jpopelka:
```suggestion
            return GithubProject(self.repo, self.service, namespace=user_login)
```

tests/integration/test_github.py
TomasTomecek: this is on purpose: previously it returned instance of GitProject and I am making sure that it really is the True value

tests/integration/test_github.py
TomasTomecek: ^

ogr/services/github.py
TomasTomecek: I missed this one, nice catch.

tests/integration/test_github.py
lachmanfrantisek: `isinstance(colin_project_fork.is_fork, bool`)?

tests/integration/test_github.py
jpopelka: me not satisfakovan
```python
    def is_forked(self) -> bool:
        return bool(self._construct_fork_project())
```

tests/integration/test_github.py
TomasTomecek: why not check directly for True?

tests/integration/test_github.py
TomasTomecek: now I don't understand, why not check directly for True?

jpopelka: no problem if you add a comment explaining it, otherwise I'd be WTF next time I see it

TomasTomecek: okay, let's compromise: let's have a full, healthy goat and also be fed

jpopelka:
I see we've confused you enough.
What I had in mind (no comment needed then) was:
```python
assert is_forked and isinstance(is_forked, bool)
```

TomasTomecek:
@jpopelka that's too simple, I prefer to overkill everything

PR #48, opened 2019-03-26
Ogr Read only  mode support with simple test

jscotka:


tests/integration/test_github_readonly.py
```diff
@@ -0,0 +1,46 @@
+import os
+
+import pytest
+from libpagure import APIError
+
+from ogr.abstract import PRStatus
+from ogr.services.mock.github import GithubService
+from tests.integration.conftest import skipif_not_all_env_vars_set
+
+pytestmark = skipif_not_all_env_vars_set(["GITHUB_TOKEN", "GITHUB_USER"])
```
TomasTomecek: I hope we won't need tokens when using the mocked interface

jscotka: Hi, sure for `full mock` environment. But For read only mode, what will be implementing `--dry-run` for `packit`, it has to be able to read data (could be possible to use anonymous access to github, but it could be unreliable)

ogr/mock_core.py
```diff
@@ -0,0 +1,93 @@
...
+from ogr.abstract import PullRequest, PRComment, PRStatus
+import datetime
+
+logger = logging.getLogger(__name__)
+
+
+class MockType(IntEnum):
+    none = 1
+    read_only = 2
+    full = 3
```
lachmanfrantisek: I would like to avoid this enum and use it like this:

- If I want `read-only` mode, I will set `read_only=True` in the constructor of the basic class like `GithubService`,`GithubProject` ...
    - We need to delegated the `read-only state to other class e.g. in the `get_project` method.

- If I want `offline`, I will use a different implementation of the ogr classes (`MockService`,`MockProject`, ...)

ogr/mock_core.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def readonly(_func=None, return_value=None, text="", use_internal_args=False):
+    def decorator_readonly(func):
+        @functools.wraps(func)
+        def readonly_func(self, *args, **kwargs):
+            read_only_suffix = "_read_only"
```
lachmanfrantisek: What about using another parameter for the function -- `read_only_function=some.alternative.function` and use it in the `read_only` mode.

This is too much magic to me...;-)

```python
# some/alternative.py
def function(self, ...):
    return something
```
```python
@readonly(read_only_function=some.alternative.function)
def some_function(self, ...):
    return something
```

But I hope, we can stay with the `return_value` and only return some fake data.

But I am not deffinitelly against the `*_read_only`...;-) (I would only remove the state for these functions.)

ogr/mock_core.py
```diff
@@ -0,0 +1,93 @@
...
+                    return getattr(self, internal_readonly_replace_name)(*args, **kwargs)
+                else:
+                    if use_internal_args:
+                        return return_value(*args, **kwargs)
+                    else:
+                        return return_value
+            elif self.mock_type == MockType.full:
+                logger.warning(f"{text} {func} with args:{args} kwargs:{kwargs}")
+
+                return MagicMock()
```
lachmanfrantisek: I hope we can avoid these `mock_type` checks as I wrote in the other comment. (And check only `self.read_only`.)

ogr/mock_core.py
jscotka: I understand your PoV, my idea to let it there was, that it could lead also to another read implementation, eg. reading via web, or something similar, to avoid any auth for `another read only` mode. 
not directly because of some `full` mocking

ogr/mock_core.py
jscotka: Why I've used this is, that I cannot pass self as parameter in decorator, What you are talking would be nice, but I have to pass also self parameter inside, what is more cryptic for me.
Super solution would be to use something like 
```
@readonly(read_only_function=self.alternative_function, pass_internal_values=True)
```
but I can use just what you said, something like what you suggested in inside decorador fuction do something like:
```
def readonly(_func,return_value=alternative_function, pass_internal_values=True, pass_object=True)
     ...
     if pass_object:
        return alternative_function(self, *args, **kwargs)
```

ogr/mock_core.py
```diff
@@ -0,0 +1,76 @@
+import functools
+import logging
+from ogr.abstract import PullRequest, PRComment, PRStatus
+import datetime
+
+logger = logging.getLogger(__name__)
+
+
+def readonly(_func=None, return_value=None, return_function=None, text=""):
```
TomasTomecek: please choose a better name of the argument "_func"; starting with underscore looks rough

TomasTomecek: also please annotate types and provide a docstring

jscotka: I've added some annotation and docs string and hope that it is good enough :-)

ogr/mock_core.py
jscotka: solved

ogr/mock_core.py
```diff
@@ -0,0 +1,93 @@
...
+             return_function: Union[Callable, None] = None,
+             text: str = "") -> Any:
+    """
+    Decorator to log  function and ovewrite return value of object methods
+    Ignore function name as first parameter and ignore every other parameters
+
+    :param return_value: returned Any value if given, return_function has higher prio if set
+    :param return_function: return function and give there parameters also
+           original caller object return_function(self, *args, **kwargs)
+    :param text: str string to put to logger output
```
TomasTomecek: why not call it log_message then? text can literally mean anything

ogr/mock_core.py
```diff
@@ -0,0 +1,93 @@
...
+from ogr.abstract import PullRequest, PRComment, PRStatus
+import datetime
+
+
+logger = logging.getLogger(__name__)
+
+
+def readonly(*,
+             return_value: Union[Any, None] = None,
+             return_function: Union[Callable, None] = None,
```
TomasTomecek: I think the preferred format is `Optional[Callable]`

jscotka: Okay, sounds reasonable. I've replaced it.

ogr/mock_core.py
```diff
@@ -0,0 +1,128 @@
...
+                if log_message:
+                    output_str += " " + log_message
+                args_str = str(args)[1:-1]
+                kwargs_str = ""
+                for k in kwargs:
+                    if isinstance(kwargs[k], str):
+                        value = "'{}'".format(kwargs[k])
+                    else:
+                        value = kwargs[k]
+                    kwargs_str += ", {}={}".format(k, value)
```
jpopelka: I think you don't need the `"'{}'".format(kwargs[k])` magic. The following should be enough.

```python
kwargs_str += f", {k}={kwargs[k]!r}"
```

```python
In [1]: s='test'

In [2]: f"{s}"
Out[2]: 'test'

In [3]: f"a {s} b"
Out[3]: 'a test b'

In [4]: f"a {s!r} b"
Out[4]: "a 'test' b"

In [5]: i=1

In [6]: f"a {i!r} b"
Out[6]: 'a 1 b'
```

jpopelka: Actually:

```python
for k, v in kwargs:
   kwargs_str += f", {k}={v!r}"
```

jscotka: ah, nice :-) thanks. another new feature what I've never used

TomasTomecek:
[test]

TomasTomecek:
@centos-ci where are you, buddy? you have some work to do [test]

TomasTomecek:
[test]

TomasTomecek:
[test] this time? (changed something in the configuration)

ogr/mock_core.py
```diff
@@ -0,0 +1,124 @@
...
+            output_str = DEFAULT_RO_PREFIX_STRING
+            if not self.read_only:
+                return func(self, *args, **kwargs)
+            else:
+                if log_message:
+                    output_str += " " + log_message
+                args_str = str(args)[1:-1]
+                kwargs_str = ""
+                for k in kwargs:
+                    kwargs_str += f", {k}={kwargs[k]!r}"
```
jpopelka: This will make `kwargs_str` start with ", ". What you want is:

```python
kwargs_str = ", ".join(f"{k}={v!r}" for k,v in kwargs.items())
```

ogr/mock_core.py
```diff
@@ -0,0 +1,124 @@
...
+        body: str,
+        commit: str = None,
+        filename: str = None,
+        row: int = None,
+    ) -> "PRComment":
+        pull_request = original_object.get_pr_info(pr_id)
+        _ = commit
+        _ = filename
+        _ = row
+        _ = pull_request
```
jpopelka: What's the purpose of these 4 lines? Is any linter complaining when you remove them?

jscotka: yep, just to not see that any arg is not used

ogr/mock_core.py
jscotka: you are right, but not sure about readability. but that's true that also current code is not very well readable :-)

jscotka:
@tomastomecek could we merge this?

TomasTomecek:
you meant @lachmanfrantisek I guess :package: 

I already stated my opinion above: LGTM

TomasTomecek:
@lachmanfrantisek that's a really good suggestion, please open a follow-up issue

PR #55, opened 2019-04-08
persistent storage for github class for testing ogr and packit

jscotka:


ogr/services/github.py
```diff
@@ -13,15 +13,23 @@
...
 from ogr.mock_core import readonly, GitProjectReadOnly
+from ogr.services.mock.github import get_Github_class
 
 logger = logging.getLogger(__name__)
 
 
 class GithubService(BaseGitService):
-    def __init__(self, token=None, read_only=False):
+    def __init__(
+        self, token=None, read_only=False, persistent_storage_file=None, ps_force=False
```
lachmanfrantisek: It takes me a while to decode the meaning of the `ps_force`.

Please, add at least a doc string for the parameters.

ogr/services/mock/pagure.py
```diff
@@ -0,0 +1,39 @@
...
+    new class method:
+        dump_yaml
+    :param storage_file: string with
+    :param write_mode: bool force write mode
+    :return: Pagure class
+    """
+    storage = PersistentObjectStorage(storage_file, write_mode=write_mode)
+    PagureMockAPI.persistent_storage = storage
+    PagureMockAPI.dump_yaml = storage.dump
+    return PagureMockAPI
```
lachmanfrantisek: Is it used anywhere?

Is it supposed to be used in the same way as the Github classes?

ogr/mock_core.py
```diff
@@ -117,3 +122,58 @@ def pr_merge(cls, original_object: Any, pr_id: int) -> "PullRequest":
     @classmethod
     def fork_create(cls, original_object: Any) -> "GitProject":
         return original_object
+
+
+class PersistentObjectStorage:
+    storage_file: str = ""
+    storage_object: dict = {}
```
TomasTomecek: this should be:
```
from typing import Dict
...
store_object: Dict
```

ogr/mock_core.py
```diff
@@ -117,3 +122,58 @@ def pr_merge(cls, original_object: Any, pr_id: int) -> "PullRequest":
...
+    write_mode: bool = False
+
+    def __init__(self, storage_file: str, write_mode: Optional[bool] = None) -> None:
```
TomasTomecek: Please document: what's write_mode?

ogr/mock_core.py
```diff
@@ -117,3 +122,58 @@ def pr_merge(cls, original_object: Any, pr_id: int) -> "PullRequest":
...
+        self.storage_file = storage_file
```
TomasTomecek: what's the semantic behind this attribute?

ogr/mock_core.py
```diff
@@ -117,3 +122,58 @@ def pr_merge(cls, original_object: Any, pr_id: int) -> "PullRequest":
...
+        else:
+            self.write_mode = not os.path.exists(self.storage_file)
+        if not self.write_mode:
+            self.load()
+
+    def __del__(self):
+        if self.write_mode:
+            self.dump()
+
+    def store(self, keys: list, values: list) -> None:
```
TomasTomecek: Please document: what does this method do?

ogr/services/mock/github.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def new__requestEncode(self, cnx, verb, url, parameters, requestHeaders, input, encode):
+    internal_keys = [verb, url, parameters]
+    if self.persistent_storage.write_mode:
+        status, responseHeaders, output = old__requestEncode(
+            self, cnx, verb, url, parameters, requestHeaders, input, encode
+        )
+        self.persistent_storage.store(
+            keys=internal_keys, values=[status, responseHeaders, output]
```
TomasTomecek: do we actually want to store headers? I'd probably just store the json payload; I'm worried that one of the headers is a session and you can get hacked easily

ogr/services/github.py
jscotka: I'll rename it to some longer name

ogr/services/mock/pagure.py
jscotka: not now, this is similar handling for pagure, I wanted to have it written just to find generic parts of code

jscotka:
> Overall, this code is a big magic, please document it.
> 
> The outcome is just awesome. I'm personally also not sure if this should be in ogr directly.

@TomasTomecek I also though where it should be. but `ogr` is I think the best place because it allows to any dependent project use these abilities (like `packit`) in very straightforward way. and it is possible direcly change behaviour just by setting init params. it is more explicit.

Second way what I though about was, to create fake gtihub module and then just via env `PYTHONPATH` change behaviour of existing project, what is also very nice and flexibile solution, but it is more hackish, because you have to know that it exist and PASS params there via env variables like these two: filenamepath to storage and if force rewrite to current file.

ogr/mock_core.py
jscotka: to have stored file what is used for reading and writing. `dump` and `load` methods uses it, to load or store data to file what is defined in contsructor. currently it is not used much, but my idea there were, that you can also explicitly call load and dump methods, whenever you want and not define eg. file parameter in  each dump method. and also it allows me, to use this file name param within __del__ to call dump when object dissappear.
Or you meant something else what I should describe

ogr/services/mock/github.py
jscotka: I also though about that and I do not know if this header is used by some method or other classes, so I copied them because I think there are no personal data what could be misused.

I understand your opinion that it can be hacked for session I also though about that and tried to find any private secure data there, that there will be session key, but I think that this session key is valid just very short time and works just for this opened session and any other session later cannot use this key. maybe it could be used as MITM attack when current session is opened in case session is right now opened.  Maybe I'm wrong

jscotka: @TomasTomecek I also looked to headers and there should not be  any private data what can be used and I think these headers are used for creating objects so I should not remove them (are forwarded by methods like: requestJsonAndCheck what calls __check method and returns it)

TomasTomecek:
[test]

TomasTomecek:
Once the CI is green, shall we merge this one?

TomasTomecek:
Newest CI run: https://ci.centos.org/job/ogr-pr/59/console yaml is not installed, please include it in the requirements.

TomasTomecek:
tests now fail :/

TomasTomecek:
Please create follow-up issue on the make target and remove the "maybe sensitive" headers from the yaml file, thank you.

PR #59, opened 2019-04-24
Implement adding PR comments

dhodovsk:
Exsmple: https://github.com/dhodovsk/rsyslog/pull/1

### DIFF
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -193,7 +193,25 @@ class GithubProject(BaseGitProject):
         filename: str = None,
         row: int = None,
     ) -> PRComment:
-        raise NotImplementedError
+        """
+        Create comment on a pull request. If creating pull request review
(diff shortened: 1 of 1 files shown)


ogr/services/github.py
```diff
@@ -174,9 +174,15 @@ def pr_comment(
         body: str,
         commit: str = None,
         filename: str = None,
-        row: int = None,
+        position: int = None,
```
TomasTomecek: please don't rename the argument because it wouldn't no longer work with other git projects:

https://github.com/packit-service/ogr/blob/master/ogr/services/pagure.py#L132

ogr/services/github.py
```diff
@@ -174,9 +174,15 @@ def pr_comment(
...
     ) -> PRComment:
```
TomasTomecek: can we also get a docstring with example values?

PR #66, opened 2019-05-02
improve mock of persistent objects

jscotka:
Improve mocking, to be easily able mock `packit` via passing `persistent_storage` class variable

### DIFF
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -40,6 +40,7 @@ class GithubService(BaseGitService):
     ):
         super().__init__()
         self._token = token
+        # it could be set as class parameter too, could be used for mocking in other projects
         if persistent_storage:
             self.persistent_storage = persistent_storage
         if self.persistent_storage:
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -21,7 +21,7 @@ from ogr.abstract import (
     CommitStatus,
 )
 from ogr.services.base import BaseGitService, BaseGitProject, BaseGitUser
-from ogr.mock_core import readonly, GitProjectReadOnly
+from ogr.mock_core import readonly, GitProjectReadOnly, PersistentObjectStorage
 from ogr.services.mock.github_mock import get_Github_class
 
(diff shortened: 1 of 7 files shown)


ogr/services/github.py
```diff
@@ -30,23 +30,22 @@
...
         self._token = token
-        if not hasattr(self, "github_class"):
-            if persistent_storage_file:
-                self.github_class = get_Github_class(
-                    persistent_storage_file, is_persistent_storage_write_mode
-                )
-            else:
-                self.github_class = github.Github
+        if persistent_storage:
+            self.persistent_storage = persistent_storage
```
lachmanfrantisek: You can avoid the `if persisent_storage` and use directly the `self.persistent_storage = persistent_storage`.

ogr/services/pagure.py
```diff
@@ -16,27 +16,26 @@
...
         self.pagure_kwargs = kwargs
-        if not hasattr(self, "pagure_class"):
-            if persistent_storage_file:
-                self.pagure_class = get_Pagure_class(
-                    persistent_storage_file, is_persistent_storage_write_mode
-                )
-            else:
-                self.pagure_class = OurPagure
+        if persistent_storage:
+            self.persistent_storage = persistent_storage
```
lachmanfrantisek: Same as above.

ogr/services/github.py
jscotka: @lachmanfrantisek No, it is not possible,
because `persistent_storage` is None by default, and I expect that tests for `packit` set this class attribute and will not pass param directly.

lachmanfrantisek: OK, it is a little confusing that it is a class attribute but we can leave it as is for now.

jscotka: yep, I can add comment there, that for mocking by other projects it is expected to use the class attribute, to be set before. because it allows you to not change current code, and just add there this change.

lachmanfrantisek: It can be confusing to overwrite the class attribute...

PR #72, opened 2019-05-22
dump() when store() is called

TomasTomecek:
\+ packit.yaml: add a link to docs

### DIFF
--- a/ogr/mock_core.py
+++ b/ogr/mock_core.py
@@ -160,13 +160,22 @@ class PersistentObjectStorage:
     is_write_mode: bool = False
     is_flushed = True
 
-    def __init__(self, storage_file: str, is_write_mode: Optional[bool] = None) -> None:
+    def __init__(
+        self,
+        storage_file: str,
+        is_write_mode: Optional[bool] = None,
+        dump_after_store: bool = False,
+    ) -> None:
         """
         :param storage_file: file name location where to write/read object data
         :param is_write_mode: force read/write mode, if not set (None) it tries to guess if
                            it should write or read data based on if file exists
+        :param dump_after_store: serialize all the data into the yaml file
+                   after calling store() - no need to call it explicitly
 
         """
+        # call dump() after store() is called
+        self.dump_after_store = dump_after_store
         self.storage_file = storage_file
         if is_write_mode is not None:
             self.is_write_mode = is_write_mode
@@ -226,12 +235,8 @@ class PersistentObjectStorage:
             current_level = current_level[item]
         self.is_flushed = False
 
-    def __del__(self):
-        try:
-            # ignore id instance deletion is done on level where is not open defined
+        if self.dump_after_store:
             self.dump()
-        except NameError:
-            pass
 
     def read(self, keys: List) -> Any:
         """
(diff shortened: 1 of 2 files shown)


ogr/mock_core.py
```diff
@@ -224,14 +222,7 @@ def store(self, keys: List, values: Any) -> None:
...
             current_level = current_level[item]
-        self.is_flushed = False
-
-    def __del__(self):
-        try:
-            # ignore id instance deletion is done on level where is not open defined
-            self.dump()
-        except NameError:
-            pass
+        self.dump()
```
jscotka: Hi @TomasTomecek I tried also this solution, but this solution were  slow in case all tests stored to one file, becasuse file were over few megabytes and it causes that each request (in store mode) stored all data and rewrite this file.

I've moved this to solution, that you have to explicitly have to call self.dump() in tearDown.

This ``def __del__(self):`` method should be removed, becase it was very unreliable. so now tests itself does not rely on that method. It causes that 90% of calls were properly stored and 10% of calls, because calling in other part of garbage collector was not able to write the data.

And using this ``is_flushed = True`` magic is also make sense to increase of performance, in case you are reading same data, it is not stored and use already stored data. to not rewrite file if nothing changed. It makes more sense in more tests using same storage file. (Both cases are not able to solve situation, when you request stores  eg. create PR, -> in sense listPR count items, then createPR, and another listPR -> count items +1) -> so that you have to use readonly mode in that case, or use more storage files, what is also possible change underlying storage file on the fly for class.

TomasTomecek: I have a problem with explicit dump calls b/c I don't have access to the PersistentObjectStorage instance. That's why I did it this way.

jscotka: @TomasTomecek where is the case you cannot access that persistent storage instance? maybe we can try to find another solution. It does not mean that I'm against this solution, just I would like to avoid it, to  not cause performance issues.

TomasTomecek: https://github.com/packit-service/packit-service/blob/master/tests/conftest.py#L32

jscotka: @TomasTomecek ah thanks, so from that perspective it is trouble when using __pure__ test methods without any class(support), that there is no setup/teardown methods (or possible but very hard to use). with stored state, just using fixtures.

So yep, if we want to use it, we can do it more explicitly if you want this:
So I suggest to change logic viceversa: so that:
``write_immediately: bool=True`` as param for persistent storage class and then use it in ``store`` method if call ``dump`` method at the end. and I'll be fine with this change

TomasTomecek: but such class doesn't solve my problem: when I want write an E2E test for packit-service, I don't have access to attributes of API, as you are setting them here:
```
        self.service_github = packit.ogr_services.GithubService(
            token=self.conf.github_token
        )
        self.project_ogr = self.service_github.get_project(
            namespace="packit-service", repo="ogr"
        )
        self.pc = get_package_config_from_repo(
            sourcegit_project=self.project_ogr, ref="master"
        )
        self.dg = packit.distgit.DistGit(self.conf, self.pc)
        self.lp = LocalProject(git_project=self.project_ogr)
        self.upstream = packit.upstream.Upstream(
            self.conf, self.pc, local_project=self.lp
        )
```

That's no use for me, I need to write a config param and hope that packit will process it.

Okay, will update this PR.

TomasTomecek:
done

TomasTomecek:
alles Grün

jpopelka:
In ordnung, fortsetzen sie bitte, mergen mergen.

Issue #80, opened 2019-06-19
Commenting on Issues


marusinm:
Based on our discussion on Issue #79, there is no functionality in ogr for commenting on Github/Pagure issues.

marusinm:
I would like to provide a solution to this issue. 

Before that, I have several questions about ogr development: 
1. when you creating such a feature like this, should one PR consists of functionality for all git forges at once. In this case, I would provide commenting on Issue for Github/Pagure/Gitlab in one PR.
2. What is current support for Gitlab? I looked at ogr/services/gitlab.py and for now, it seems to be unsupported.
3.  Is there any special way how to run tests? Or something I should know about writing them? *Release-bot*, for example, creates its own container for testing with a makefile.

rpitonak:
> 1. when you creating such a feature like this, should one PR consists of functionality for all git forges at once. In this case, I would provide commenting on Issue for Github/Pagure/Gitlab in one PR.

I think this is up to you, small PRs are easier to review. About Gitlab, as far as I know, we focus mainly on Github and Pagure right now.

> 3. Is there any special way how to run tests? Or something I should know about writing them? _Release-bot_, for example, creates its own container for testing with a makefile.

@jscotka can provide the best answer to this question.

TomasTomecek:
> I would like to provide a solution to this issue. 
> 
> Before that, I have several questions about ogr development: 
> 
>     1. when you creating such a feature like this, should one PR consists of functionality for all git forges at once. In this case, I would provide commenting on Issue for Github/Pagure/Gitlab in one PR.

Since none of us is really using the gitlab interface, I wouldn't worry with it much (unless you like adventure).

As Rado said, send the PR as you wish. One PR is okay.

>     2. What is current support for Gitlab? I looked at ogr/services/gitlab.py and for now, it seems to be unsupported.

Yup, no one is using it (yet).

>     3. Is there any special way how to run tests? Or something I should know about writing them? _Release-bot_, for example, creates its own container for testing with a makefile.

https://github.com/packit-service/ogr/blob/master/CONTRIBUTING.md#makefile

please let us know if the doc is outdated

marusinm:
I successfully run tests on my local machine, now I wanted to test my new functionality for this Issue. 

However, I need to generate a similar file like .yaml files stored in tests/integration/test-data. Without that, I am not able to test new functionality. Any advice on how to generate such .yaml file?

jscotka:
[...]

PR #99, opened 2019-06-30
creating/closing/commenting Pagure Issues

marusinm:
Related to #80

This PR contains functionality for Pagure issues except for tagging issues. I had some problems with tagging, therefore, I'll create separated PR for it.

If I am right I have to set Pagure token for creating yaml test-files. We can manage that in private chat, then I'll add tests.

### DIFF
--- a/ogr/services/pagure.py
+++ b/ogr/services/pagure.py
@@ -350,8 +350,8 @@ class PagureProject(BaseGitProject):
 
     def issue_close(self, issue_id: int) -> Issue:
         payload = {"status": "Closed"}
-        issue = self.get_issue_info(issue_id)
         self._call_project_api("issue", str(issue_id), "status", data=payload, method="POST")
+        issue = self.get_issue_info(issue_id)
         return issue
 
     def get_pr_list(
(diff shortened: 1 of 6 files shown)


ogr/services/pagure.py
```diff
@@ -512,6 +549,45 @@ def get_git_urls(self) -> Dict[str, str]:
...
+        else:
+            url = "/".join(
+                [
+                    self.service.instance_url,
+                    self.namespace,
+                    self.repo,
+                    "issue",
+                    str(issue_dict["id"]),
+                ]
+            )
```
lachmanfrantisek: Try using [self.getproject_url](https://github.com/packit-service/ogr/blob/0d60bcc68e53db7d75cb898c036c7b4c529f045a/ogr/services/pagure.py#L296)...

This should work for you:
```python
url = self.get_project_url("issue", str(issue_dict["id"]))
```

ogr/services/pagure.py
```diff
@@ -317,6 +317,43 @@ def get_branches(self) -> List[str]:
...
+        payload = {
+            "title": title,
+            "issue_content": body
+        }
+        new_issue = self._call_project_api("new_issue", data=payload, method="POST")["issue"]
+        return self._issue_from_pagure_dict(new_issue)
+
+    def issue_close(self, issue_id: int) -> Issue:
+        payload = {"status": "Closed"}
+        issue = self.get_issue_info(issue_id)
```
lachmanfrantisek: Do we need this call?

lachmanfrantisek: Oh, I see that you would like to return it and we cannot get info from the response...
(We should create an issue on Pagure for adding issue-info to the response.)

The status of the issue would be outdated in this order -- place this call after the "changing" call.

Issue #100, opened 2019-07-05
user's permissions on Issues/PRs


marusinm:
ogr.abstract.Issue and ogr.abstract.PullRequest doesn't contain a field if a user has permission to close Issue or merge PR. 

It should be detected by the user's token. In the case of Github, roles of `owner`, `member`, `collaborator` should have permissions. 

It could be related to #88 if we save a list of owners, members, and collaborators into GitProject instance or create get methods for them.

lachmanfrantisek:
Isn't this already fixed for GitHub and Pagure?

lachmanfrantisek:
- We have it in Gitlab as well.
- We have `can_merge_pr`/`who_can_merge_pr`/`who_can_close_issue` methods in the project classes.

Can we close this issue?

Or, do we want/need related methods in Issue/PR classes as well? (`my_issue.who_can_close`/`issue_can_close("username")`)

lachmanfrantisek:
I've updated the description. This issue is now about cleaning the methods and add the missing parts. Please, check if I don't forget anything.

pranav33317:
Yes ma'am , I am working on it , will draft a PR ASAP . Had a question about the unimplemented functions , a lot of the functions are missing return statements and thus giving error . Should I just return default value or raise some error ?
Thanks

lbarcziova:
@pranav33317 hello and thanks! 

> a lot of the functions are missing return statements and thus giving error .

Could you please provide some examples?

pranav33317:
def get_issue_list(
        self,
        status: IssueStatus = IssueStatus.open,
        author: Optional[str] = None,
        assignee: Optional[str] = None,
        labels: Optional[list[str]] = None,
    ) -> list[Issue]:
        pass

    @indirect(GithubIssue.get)
    def get_issue(self, issue_id: int) -> Issue:
        pass

    @indirect(GithubIssue.create)
    def create_issue(
        self,
        title: str,
        body: str,
        private: Optional[bool] = None,
        labels: Optional[list[str]] = None,
        assignees: Optional[list[str]] = None,
    ) -> Issue:
        pass , @indirect(GithubPullRequest.get_list)
    def get_pr_list(self, status: PRStatus = PRStatus.open) -> list[PullRequest]:
        pass

    @indirect(GithubPullRequest.get)
    def get_pr(self, pr_id: int) -> PullRequest:
        pass , def create_pr(
        self,
        title: str,
        body: str,
        target_branch: str,
        source_branch: str,
[...]

Issue #88, opened 2019-06-26
Add get_owner/get_owners method to project classes


lachmanfrantisek:
It would be useful to have a method for getting the owner(s) of the project.

If it is possible to have multiple owners, create `get_owners` and return list of strings.

shreyanshrs44:
Hi, I would like to work on this issue. From where I should start to work on this?

lachmanfrantisek:
It would be nice!

- In case of Pagure, you need to get the info from the API:
    - every instance has a documentation on /api/0/ (e.g. https://src.fedoraproject.org/api/0/)
    - the relevant would be part about `Project information`
    - since we are using this info in multiple was, we already have a function for that: [get_project_info](https://github.com/packit-service/ogr/blob/master/ogr/services/pagure.py#L309).

- In case of Github, we are working with pygithub objects:
    - you have the `Repository` instance in `self.github_repo`
    - [documentation for the class](https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html#github.Repository.Repository.owner) (We need to return string, not the instance of [NamedUser](https://pygithub.readthedocs.io/en/latest/github_objects/NamedUser.html). The `login` is the attribute we want.)


- Please, add also the method to the [GitProject](https://github.com/packit-service/ogr/blob/d1f683b456c87b19720fe8a0b8b71061b65642b0/ogr/abstract.py#L280).

marusinm:
@shreyanshrs44 are you currently working on this issue? I opened #88 which includes also feature in this issue. I would provide a solution in one PR.

marusinm:
@shreyanshrs44 meanwhile I implemented this in #101 since I need this feature in another project.

> From where I should start to work on this?

If you'll refactor my code It may be a nice showcase on how to start work on such issues.

lachmanfrantisek:
Already fixed in #101

PR #101, opened 2019-07-08
get project's owners and permissions of various users

marusinm:
Implementation of getting project owners related to #88. 
And getting users with permission for closing Issues and merging PR based on [Github permissions](https://help.github.com/en/articles/repository-permission-levels-for-an-organization) and [Pagure permissions](https://docs.pagure.org/pagure/usage/project_acls.html) related to issue #100.

### DIFF
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -229,6 +229,42 @@ class GithubProject(BaseGitProject):
                 return None
         return self._construct_fork_project()
 
+    def get_owners(self) -> List[str]:
+        # in case of github, repository has only one owner
+        owners = []
+        owner = self.github_repo.owner
+        owners.append(owner.login)
+        return owners
+
+    def who_can_close_issue(self) -> List[str]:
+        usernames = []
+        collaborators = self._get_collaborators_with_permission()
+        for username, persmission in collaborators.items():
+            if persmission == "admin" or persmission == "write":
+                usernames.append(username)
+        return usernames
+
+    def who_can_merge_pr(self) -> List[str]:
+        usernames = []
+        collaborators = self._get_collaborators_with_permission()
+        for username, persmission in collaborators.items():
+            if persmission == "admin" or persmission == "write":
+                usernames.append(username)
+        return usernames
+
+    def _get_collaborators_with_permission(self) -> dict:
+        """
+        Get all project collaborators in dictionary with permission association
+        :return: List of usernames
+        """
+        collaborators = {}
+        users = self.github_repo.get_collaborators()
+        for user in users:
+            permission = self.github_repo.get_collaborator_permission(user)
+            collaborators[user.login] = permission
+            # collaborators.append(user.login)
+        return collaborators
+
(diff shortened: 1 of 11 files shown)


ogr/services/pagure.py
```diff
@@ -357,6 +357,27 @@ def get_branches(self) -> List[str]:
     def get_description(self) -> str:
         return self.get_project_info()["description"]
 
+    def get_owners(self) -> List[str]:
+        project = self.get_project_info()
+        return project["access_users"]["owner"]
+
+    def who_can_close_issue(self) -> List[str]:
+        users: Set[str] = set()
```
lachmanfrantisek: What about returning `set` for all methods (github/pagure/abstract)?

marusinm:
>Isn't it possible to close an issue created by myself even if I cannot do that on other issues?

@lachmanfrantisek I completely changed approach from previous commit how to get user permissions. For example, I changed fnc `who_can_close_issue() -> List[users]` to ` can_close_issue(user, issue) -> bool`. Using this new functions should be more straightforward.

ogr/services/github.py
```diff
@@ -228,6 +228,55 @@ def get_fork(self, create: bool = True) -> Optional["GithubProject"]:
...
+    def can_close_issue(self, username: str, issue: Issue) -> bool:
+        try:
+            collaborators = self._get_collaborators_with_permission()
+        except github.GithubException:
+            logger.debug(
+                f"{username} must have push access to view repository permissions."
+            )
+            return False
+
+        for login, permission in collaborators.items():
```
lachmanfrantisek: Personally, I would like to see also the list/set of allowed users. (Since you are going through all of the collaborators and we are not saving any time with that.)

We have following possibilities:

1. `who_can_close_issues` and `can_close_issue` (which will use `who_can_close_issues` and checks the author of the issue) (We can move the `can_close_issue` to the Issue object later.)
2. `who_can_close_issue` with the optional parameter for specifying the issue id
3. leave it in the current state

I vote for the second one but make your own decision...;-)

(Same for the pull-requests.)

ogr/services/github.py
```diff
@@ -229,6 +229,74 @@ def get_fork(self, create: bool = True) -> Optional["GithubProject"]:
...
+                usernames.append(login)
+
+        return set(usernames)
+
+    def can_close_issue(self, username: str, issue: Issue) -> bool:
+        allowed_users = self.who_can_close_issue()
+
+        for allowed_user in allowed_users:
+            if username == allowed_user:
+                return True
```
lachmanfrantisek: What about this:

```python
if username in allowed_users:
    return True
```

ogr/services/github.py
```diff
@@ -229,6 +229,74 @@ def get_fork(self, create: bool = True) -> Optional["GithubProject"]:
...
+            return True
+
+        return False
+
+    def can_merge_pr(self, username) -> bool:
+        allowed_users = self.who_can_merge_pr()
+
+        for allowed_user in allowed_users:
+            if username == allowed_user:
+                return True
```
lachmanfrantisek: Same as above.

PR #114, opened 2019-07-11
method create release for github created

### LINKED ISSUE #111: Allow creating releases
For `Github` we can get the releases:
https://github.com/packit-service/ogr/blob/master/ogr/services/github.py#L571

We need to also create a new release from API.

---

Relevant documentation:
https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html#github.Repository.Repository.create_git_release

lbarcziova:
Fixes #111

### DIFF
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -29,6 +29,7 @@ from github import (
     IssueComment as GithubIssueComment,
     Repository,
     CommitComment as GithubCommitComment,
+    GithubObject,
 )
 from github.GitRelease import GitRelease as GithubRelease
 from github.Issue import Issue as GithubIssue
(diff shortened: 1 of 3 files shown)


tests/integration/test_github.py
```diff
@@ -111,6 +114,14 @@ def test_get_releases(self):
 
         assert len(releases) >= 9
 
+    def test_create_release(self):
+        release = self.hello_world_project.create_release(
+            tag="0.3.0", name="test", message="testing release"
+        )
+        assert release.tag_name == "0.3.0"
+        assert release.title == "test"
+        assert release.body == "testing release"
```
lachmanfrantisek: Maybe, we can test also the increase in the number of releases. (`before + 1 == after`)

ogr/services/github.py
```diff
@@ -595,6 +596,25 @@ def get_releases(self) -> List[Release]:
...
         ]
 
+    def create_release(
+        self,
+        tag: str,
+        name: str,
+        message: str,
+        draft=False,
+        prerelease=False,
+        target_commitish=GithubObject.NotSet,
```
lachmanfrantisek: I would go with the subset of properties that can be implemented also in the other implementation.

(It would be nice to avoid Github specific objects in the signature of the method...)

PR #119, opened 2019-07-15
unused functions in utils.py removed

### LINKED ISSUE #103: Clean the ogr/utils.py
In #102, I've realized that most of the functions in [utils.py](https://github.com/packit-service/ogr/blob/master/ogr/utils.py) are never used.

We need to go through the functions and check if they are not used anywhere (please, control the packit codebase as well) and remove the unused ones.

----

In the other hand, we can move here the [LocalProject](https://github.com/packit-service/packit/blob/master/packit/local_project.py) or other git-related code from the Packit. Do we want OGR to be only remote API or git-helper?

lbarcziova:
Fixes #103

### DIFF
--- a/ogr/utils.py
+++ b/ogr/utils.py
@@ -20,12 +20,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import datetime
 import logging
 import os
 import re
 import subprocess
-import tempfile
 from typing import List, Union, Match, Optional
 
 import six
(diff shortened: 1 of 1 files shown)


lachmanfrantisek:
I've added the following question to the original issue:

> In the other hand, we can move here the LocalProject or other git-related code from the Packit. Do we want OGR to be only remote API or git-helper?

I think we need to have a common agreement to that before merging this.

However, we can remove this now and move/merge the git-related code later.

TomasTomecek:
We should talk about that first honestly, I would not try to overuse this PR.

jpopelka:
Summary from a discussion:
- merge this
- spike (RHELPLAN-24239) high-level library (ogr?) for similar git-related utils scattered in our projects

PR #128, opened 2019-07-18
add is_merged field into PullRequest

marusinm:
`PullRequest` class in abstract.py already has `status` field. 

In the case of Github, the set of values from API for this field is `open`/`closed`/`all` and therefore there is missing value `merged`. I added `is_merged` field into `PullRequest` class (since it causes bugs in release-bot).

I set the new `is_merged` field also for Pagure.


marusinm:
> Since we have a custom enum for PRStatus, I would rather go with extending that enum with another value.

Yes, we have `PRStatus`. Actually, there is already value `merged` but the behavior between Github and Pagure is different. 

after call `project.get_pr_list(PRStatus.merged)`
- Pagure returns all merged PRs
- Github returns empty list because PRStatus.merged doesn't exist in Github's API


after call `project.get_pr_list(PRStatus.closed)`
- Pagure returns closed PRs (not merged)
- Github returns closed and merged PRs together. Ogr parse it into `PullRequest` class and we cannot recognize whether PR is merged or just closed, that's why I added `is_merged` field into this class.

I can change it but I'm not sure about another solution.

lachmanfrantisek:
@marusinm thanks for the investigation.

I think that Gitlab and Pagure use the same way -- `merged` is one of the states.

I would vote for fixing the GitHub responses to be consistent with Pagure and Gitlab.

lachmanfrantisek:
Let's close this one in favour of #130

Issue #126, opened 2019-07-18
get_email for GitUser


marusinm:
We have `get_username` method in GitUser class. At the moment we are missing method `get_email` for both ~~Github~~ and Pagure services.

-----

- [x] GitHub
- [x] GitLab
- [ ] Pagure

yzhang2907:
@marusinm I added the method, will submit PR now. Thank you!

lachmanfrantisek:
Pagure still misses that.

lachmanfrantisek:
Looks like we are missing only Pagure know which does not provide that value.

See: https://src.fedoraproject.org/api/0/#users-tab

lachmanfrantisek:
Can anyone check that, please?

We have two options:
- Create a bug report (since email can be kind of private information, I am not sure if they will be willing to do that) and implement that
- Raise some meaningful exception for Pagure and close this issue. (Preffered by me..)

csomh:
Would it be possible to return `None` for Pagure? I feel an exception would be a little bit harsh.

mfocko:
will resolve with Friday work

PR #129, opened 2019-07-18
Fix Issue #126, implemented get_email

yzhang2907:


lachmanfrantisek:
[test]

marusinm:
Also, [there is a test](https://github.com/packit-service/ogr/blob/11bec2cb6fd9cc4a0f707078db55e6c09f2d6b75/tests/integration/test_github.py#L128) for `get_username` method. You can use something similar and simple as well.

rpitonak:
> In the case of Pagure, we are not using any library, but calling the API on our own. Look at the other methods, how they are implemented in `ogr/services/github.py`. The API documentation of pagure can be seen e.g. at https://src.fedoraproject.org/api/0/

Just a small correction, here should be probably `ogr/services/pagure.py` instead of  `ogr/services/github.py`.

yzhang2907:
@lachmanfrantisek @marusinm Thank you! I will get it fixed today.

yzhang2907:
I cannot make check locally on either Windows or Mac OS machine; I think it has to be a Fedora distro. Sorry lads, tried my best :(

lachmanfrantisek:
> I cannot make check locally on either Windows or Mac OS machine; I think it has to be a Fedora distro.

It should not be dependant on the Fedora. I've described the testing in the review which can solve the problem.

In the case of [pre-commit](https://pre-commit.com/), you can find better log output [here](https://softwarefactory-project.io/logs/29/129/eaa3587168e64752c811aafb3fb42c7be80a8b8d/check/pre-commit/501b5cb/log-classify.html).

It's a group of static-analysis tools/checks and it finds some problems in your code..;-)

- Two things can be fixed by my suggestions from review.
- The missing problem is `ogr/services/pagure.py:840:5: E303 too many blank lines (2)`

tests/integration/test_github.py
```diff
@@ -129,6 +129,13 @@ def test_username(self):
         # changed to check just lenght, because it is based who regenerated data files
         assert len(self.service.user.get_username()) > 3
 
+    def test_email(self):
+        test_str = self.service.user.get_email()
+        assert test_str != None
```
lachmanfrantisek:
```suggestion
        assert test_str
```
This should be enough, and you will also fix the problem found by `pre-commit`.

tests/integration/test_pagure.py
```diff
@@ -129,6 +129,13 @@ def test_username(self):
         # changed to check just lenght, because it is based who regenerated data files
         assert len(self.service.user.get_username()) > 3
 
+    def test_email(self):
+        test_str = self.service.user.get_email()
+        assert test_str != None
```
lachmanfrantisek:
```suggestion
        assert test_str
```

Same as above.

yzhang2907:
Sorry for the late response, I just saw this, I will be working on it! Thank you for being patient with me!!

lachmanfrantisek:
> Sorry for the late response, I just saw this, I will be working on it! Thank you for being patient with me!!

No problem, feel free to ask if you need some help..;-)

ogr/services/pagure.py
```diff
@@ -824,6 +824,12 @@ def get_username(self) -> str:
         return_value = self.service.call_api(url=request_url, method="POST", data={})
         return return_value["username"]
 
+    def get_email(self) -> str:
+        request_url = self.service.get_api_url("user", self.get_username())
+
+        return_value = self.service.call_api(url=request_url)["default_email"]
```
marusinm: This addition won't work since Pagure API doesn't support `default_email` parameter in your API call on `GET /api/0/user/<username>`. I just created PR #141 and put your working commits there.

rpitonak:
@yzhang2907 thank you for your contribution! You commits were included in the follow-up [PR](https://github.com/packit-service/ogr/pull/141) from @marusinm which has been merged recently.

PR #131, opened 2019-07-21
add get_latest_release(), create Releases class in test_github.py

marusinm:
It's pretty common looking for the latest release only. It would be useful to have a fast method for it without looking for some identifier or getting all releases. Here it is. 

I also think that tests for releases were in the incorrect class. Release tests which are increasing deserve its own class.


lachmanfrantisek:
@marusinm Could you please rebase to the master?

PR #136, opened 2019-07-25
run tests on one repository

lbarcziova:
Fixes #124

Issue #86, opened 2019-06-25
object-specific methods bound to ogr classes


lachmanfrantisek:
- [x] Forward some object to the issue/pr/... instances to allow connection to the service. (Classes are not only data-holders.)
- [x] All ogr classes to have properties with setters allowing to update its attributes
    - I can do `my_pr.title = "something"`
- [x] Objects are linked together (Service <-> Project <-> PR <-> PRComment)
- [x] Keep old methods, mark them as deprecated.

needs #121

mfocko:
- [ ] Prepare file structure
- [ ] Comments

lachmanfrantisek:
@mfocko What's needed for the `Release`? Only the setters for properties?

mfocko:
> @mfocko What's needed for the `Release`? Only the setters for properties?

- method to create object from `raw_release`
- fix tests
- looking at the code probably move some things from abstract to base class

lachmanfrantisek:
Thanks for the update

PR #137, opened 2019-07-25
add method edit_release

### LINKED ISSUE #127: edit Github release
Can we create a method for editing a release? We can use [this](https://pygithub.readthedocs.io/en/latest/github_objects/GitRelease.html#github.GitRelease.GitRelease.update_release) for it.

lbarcziova:
Fixes #127

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -220,6 +220,7 @@ class Release:
         created_at: str,
         tarball_url: str,
         git_tag: GitTag,
+        project=None,
     ) -> None:
         self.title = title
         self.body = body
@@ -228,6 +229,7 @@ class Release:
         self.created_at = created_at
         self.tarball_url = tarball_url
         self.git_tag = git_tag
+        self.project = project
 
     def save_archive(self, filename):
         response = urlopen(self.tarball_url)
@@ -247,6 +249,15 @@ class Release:
             f"created_at='{self.created_at}',"
             f"tarball_url='{self.tarball_url}')"
         )
+
+    def edit_release(self, name: str, message: str):
+        """
+        Edit name and message of a release.
+
+        :param name: str
+        :param message: str
+        """
+        raise NotImplementedError()
 
 
 class GitService:
@@ -530,11 +541,13 @@ class GitProject:
         """
         raise NotImplementedError()
 
-    def get_release(self, identifier: int) -> Release:
+    def get_release(self, identifier=None, name=None, tag_name=None) -> Release:
         """
         Get a single release
 
-        :param identifier:
+        :param identifier: int
+        :param name: str
+        :param tag_name: str
         :return: Release
         """
         raise NotImplementedError()
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -218,7 +218,7 @@ class Release:
         created_at: str,
         tarball_url: str,
         git_tag: GitTag,
-        project: Optional["GitProject"],
+        project: "GitProject",
     ) -> None:
         self.tag_name = tag_name
         self.url = url
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -213,23 +213,27 @@ class GitTag:
 class Release:
     def __init__(
         self,
-        title: str,
-        body: str,
         tag_name: str,
         url: str,
         created_at: str,
         tarball_url: str,
         git_tag: GitTag,
-        project=None,
+        project: Optional["GitProject"],
     ) -> None:
-        self.title = title
-        self.body = body
         self.tag_name = tag_name
         self.url = url
         self.created_at = created_at
         self.tarball_url = tarball_url
         self.git_tag = git_tag
         self.project = project
+
+    @property
+    def title(self):
+        raise NotImplementedError()
+
+    @property
+    def body(self):
+        raise NotImplementedError()
 
     def save_archive(self, filename):
         response = urlopen(self.tarball_url)
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -30,7 +30,7 @@ from github import (
     Repository,
     CommitComment as GithubCommitComment,
 )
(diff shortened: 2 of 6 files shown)


ogr/services/github.py
```diff
@@ -721,6 +721,21 @@ def create_release(self, tag: str, name: str, message: str) -> Release:
         )
         return self.get_release(created_release.id)
 
+    def edit_release(self, identifier: int, name: str, message: str):
```
lachmanfrantisek: I am not sure if the identifier is the right input. (Also in the `get_release` method since it is some internal number, and we do not have it even in the [Release class](https://github.com/packit-service/ogr/blob/master/ogr/abstract.py#L213).)
I am also not sure if this is something we have in the other git forges (Pagure, Gitlab).

I know it would not be as easy as with id, but we need to get releases by tag (ideally) or by name.

What about the following methods:

- `get_releases() -> Release`
- `get_last_release() -> List[Release]`
- `get_release(identifier=None, name=None, tag=None) -> Release` (you need to set one and only one, leave the identifier for backward compatibility)
- I do not know what would be a best way to update release. What about adding it directly to `Release` class:

- [ ] Make a `GithubRelease` as a subclass of `Release` which will contain release representation of PyGithub.
- [ ] Create `GithubRelease` instances instead of `Release` in `GithubProject`.
- [ ] Add the `edit_release` as an abstract method to the `Release` class and implement it in the `GithubRelease`.

Anyway, we have a plan in #86 to implement all Issue/PR/Release related methods directly in those objects, so why not start with this now.

ogr/abstract.py
```diff
@@ -220,6 +220,7 @@ def __init__(
         created_at: str,
         tarball_url: str,
         git_tag: GitTag,
+        project=None,
```
lachmanfrantisek:
```suggestion
        project: Optional["GitProject"]=None,
```

lachmanfrantisek:
```suggestion
        project: Optional["GitProject"],
```

Maybe, we can make it compulsory. What do you think?

ogr/services/github.py
```diff
@@ -57,6 +57,38 @@
...
+    def edit_release(self, name: str, message: str) -> None:
+        """
+        Edit name and message of a release.
+
+        :param name: str
+        :param message: str
+        """
+        self.raw_release.update_release(name=name, message=message)
+        self.title = self.raw_release.title
+        self.body = self.raw_release.body
```
lachmanfrantisek: I have just one idea about this. It would be usefull to use properties for that so you do not need to update the attributes:

```suggestion
        
    @property
    def body(self):
        return self.raw_release.body
```

(Same for the `self.title` and other attributes.)

I am just worried, that we need to remove these atributes from the `__init__` method of the `Release` and create abstract properties for each of them in the `Release` instead:

```python
class Release:

    def __init__(self):
        pass

    @property
    def body(self):
        raise NotImplementedError()
```

Not sure if it is worth it...

ogr/services/github.py
```diff
@@ -693,13 +729,31 @@ def _normalize_label_color(color):
             return color[1:]
         return color
 
-    def get_release(self, identifier: int) -> Release:
+    def release_id_from_name(self, name):
```
lachmanfrantisek: You can use `_release_id_from_name` instead to say that it is not a part of our API and we are using that method only internally.

ogr/services/github.py
```diff
@@ -693,13 +729,31 @@ def _normalize_label_color(color):
...
+        releases = self.github_repo.get_releases()
+        for release in releases:
+            if release.title == name:
+                return release.id
+        return None
```
lachmanfrantisek: Please, add type annotation with `Optional` to indicate that we can get `None` as well.

```python
def release_id_from_name(self, name) -> Optional[Release]:
```

ogr/services/github.py
```diff
@@ -693,13 +729,31 @@ def _normalize_label_color(color):
...
+
+    def release_id_from_tag(self, tag):
+        releases = self.github_repo.get_releases()
+        for release in releases:
+            if release.tag_name == tag:
+                return release.id
+        return None
```
lachmanfrantisek: Same note about the name with `_` and type as above.

ogr/services/github.py
```diff
@@ -693,13 +729,31 @@ def _normalize_label_color(color):
...
+
+    def get_release(self, identifier=None, name=None, tag_name=None) -> GithubRelease:
+        if tag_name:
+            identifier = self.release_id_from_tag(tag_name)
+        elif name:
+            identifier = self.release_id_from_name(name)
```
lachmanfrantisek: Maybe some exception if we did not find anything? (Or we got `None` for all.)

tests/integration/test_github.py
```diff
@@ -282,6 +287,19 @@ def test_create_release(self):
         assert release.body == "testing release"
         assert count_before + 1 == count_after
 
+    def test_edit_release(self):
+        release = self.hello_world_project.get_release(tag_name="0.1.0")
+        origin_name = release.title
+        origin_message = release.body
+
+        release.edit_release(name="changed name", message="edited message")
```
lachmanfrantisek:
```suggestion
        release.edit_release(name=f"{origin_name}-changed", message=f"{origin_message}-changed")
```

Just to be sure we change anything...

ogr/abstract.py
```diff
@@ -213,21 +213,27 @@ def __str__(self) -> str:
...
     def __init__(
         self,
-        title: str,
-        body: str,
         tag_name: str,
         url: str,
         created_at: str,
         tarball_url: str,
         git_tag: GitTag,
+        project: Optional["GitProject"],
```
lachmanfrantisek:
```suggestion
        project: "GitProject",
```

It is not needed now...

lachmanfrantisek:
@lbarcziova The [second PR](#136) was merged. Could you please rebase to be sure it really works together? Thanks!

PR #140, opened 2019-08-05
Function for getting all commits from specific PR.

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

Pull request adds a function for getting all commits for specific PR.
It is useful when the user adds a comment into `packit-service` like `/packit copr-build` and we would like to get especially the latest commit from the PR.

A real example:
https://github.com/packit-service/packit-service/blob/master/packit_service/worker/github_handlers.py#L92 `pr_event/base_ref` cannot by `pr/9` but reference to the latest commit for a pull request.

- [x] Pull Request contains info about SHA of the latest commit.
- [x] tests are covered.


ogr/services/github.py
```diff
@@ -395,6 +395,10 @@ def get_pr_info(self, pr_id: int) -> PullRequest:
         pr = self.github_repo.get_pull(number=pr_id)
         return self._pr_from_github_object(pr)
 
+    def get_all_pr_commits(self, pr_id: int) -> List[str]:
+        pr = self.github_repo.get_pull(number=pr_id)
+        return list([commit.sha for commit in pr.get_commits()])
```
jpopelka:
```suggestion
        return [commit.sha for commit in pr.get_commits()]
```

tests/integration/test_github.py
```diff
@@ -239,6 +238,20 @@ def test_pr_info(self):
         assert pr_info.title.startswith("new")
         assert pr_info.status == PRStatus.merged
 
+    def test_all_pr_commits(self):
+        commits = self.colin_project.get_all_pr_commits(pr_id=1)
+        assert commits
+        assert len(commits) == 7
+        assert commits == [
```
jpopelka: I think you don't need the first two asserts since the third implicates them.

ogr/services/github.py
phracek: Good catch. I forgot, that PaginatedList is related only for `pr.get_commit()`.

~~~
In [1]: from ogr import GithubService
In [2]: service = GithubService(token="e620d8c9979385e6b0764b91c9467c6447585519")
In [3]: colin = service.get_project(namespace="user-cont", repo="colin")
In [4]: pr = colin.github_repo.get_pull(number=1)
In [5]: type([commit.sha for commit in pr.get_commits()])
Out[5]: list
In [6]: type(pr.get_commits())
Out[6]: github.PaginatedList.PaginatedList
In [7]:
~~~

tests/integration/test_github.py
phracek: You are right with `len(commits)` but in whole `tests` you can see similar tests.
https://github.com/packit-service/ogr/blob/master/tests/integration/test_github.py#L62 https://github.com/packit-service/ogr/blob/master/tests/integration/test_github.py#L240

Then I would prefer to rewrite `test_github.py` so we have the same or similar tests code structure.

PR #144, opened 2019-08-08
Add get_pr_commits into abstract.py

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

Add function `get_pr_commits` into `abstract.py`.

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -579,7 +579,7 @@ class GitProject:
 
     def get_pr_commits(self, pr_id: int) -> List[str]:
         """
-        Get list of pull-request commits.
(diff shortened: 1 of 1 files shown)


ogr/abstract.py
```diff
@@ -577,6 +577,15 @@ def get_pr_comments(
         """
         raise NotImplementedError()
 
+    def get_pr_commits(self, pr_id: int) -> List[str]:
+        """
+        Get list of pull-request commits.
+
+        :param pr_id: int
+        :return: [str]
```
jpopelka: I guess the `str` means the commit hash, but you might make that explicit. Thanks.

PR #155, opened 2019-08-15
Run zuul for pip and rpm

lachmanfrantisek:
- Run zuul tests both on pip and rpm.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -3,17 +3,21 @@
     name: packit-service/ogr
     check:
       jobs:
-        - tests
-        - pre-commit
+        - ogr-pre-commit
+        - ogr-tests-rpm
+        - ogr-tests-pip-deps
     gate:
       jobs:
-        - tests
-        - pre-commit
+        - ogr-pre-commit
+        - ogr-tests-rpm
+        - ogr-tests-pip-deps
+
 - job:
-    name: tests
+    name: ogr-tests-rpm
     parent: base
-    description: Run tests of ogr
-    run: recipe.yaml
+    description: Run tests of packit via rpms
+    pre-run: files/zuul-install-requirements-rpms.yaml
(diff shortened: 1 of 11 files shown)


.zuul.yaml
```diff
@@ -3,26 +3,46 @@
     name: packit-service/ogr
     check:
       jobs:
-        - tests
-        - pre-commit
+        - ogr-pre-commit
```
rpitonak: wouldn't `pre-commit` be enough?

.zuul.yaml
```diff
@@ -3,26 +3,46 @@
...
+        - ogr-tests-rpm
+        - ogr-tests-pip-deps
     gate:
       jobs:
-        - tests
-        - pre-commit
+        - ogr-pre-commit
```
rpitonak: same as above

files/tasks/build-rpm-deps.yaml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Install build RPM dependencies
```
rpitonak: I would switch `RPM` with `build`

.zuul.yaml
lachmanfrantisek: The zuul needs identical job names accross all projects.

As you can see in the first comment by zull that is hidden:

```
Job tests-rpm in packit-service/packit is not permitted to shadow job tests-rpm in packit-service/ogr
```

So I added prefix to all jobs.

files/tasks/build-rpm-deps.yaml
lachmanfrantisek: I've copied that from packit but I'll change it.

.zuul.yaml
rpitonak: Ok, I see. So this is resolved then.

Issue #159, opened 2019-08-20
Add method for loading instances from dict


lachmanfrantisek:
In [packit](https://github.com/packit-service/packit) and [upsint](https://github.com/packit-service/upsint/), we need to load authentication for services from files.

It would be nice to have a method for loading instances from `dict` so we can use it from multiple projects.

```python
>>> from ogr import get_instances_from_dict
>>> get_instances_from_dict({
...    "github.com": {"token": "abcd"},
...    "pagure.io": {"token": "defg"},
...    "src.fedoraproject.org": {"token": "asdasdasda"},
... })
[
    GithubService(token="abcd"),
    PagureService(instance_url="https://pagure.io", token="defg"),
    PagureService(instance_url="https://src.fedoraproject.org", token="asdasdasda"),
]
```

TomasTomecek:
would be nice to see the dict definition before working on an implementation

lachmanfrantisek:
> would be nice to see the dict definition before working on an implementation

I've proposed one in the description.

key: "server_url"
value: dict with arguments passed to `__init__`

lachmanfrantisek:
This approach is good for most use-cases but does not allow custom mapping. (When the mapping url->servise-type is not known to us.)

We can add some optional argument for that:

```python
{
    "my.gtlb": {
        "type": "gitlab",
        "token": "abcd",
        "ssl_verify" : False
    }
}
```

lachmanfrantisek:
@TomasTomecek Feel free to discuss the format in the PR: #160

PR #160, opened 2019-08-21
Loading services from dict

lachmanfrantisek:
- Add method for loading services from dictionary.
- Tests included.

Fixes: #159

### DIFF
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 
 import functools
-from typing import Dict, Type, List
+from typing import Dict, Type, List, Optional, Set
 
 from ogr.abstract import GitService, GitProject
 from ogr.exceptions import OgrException
@@ -99,6 +99,28 @@ def get_project(
     return project
 
 
+def get_service_class_or_none(
+    url: str, service_mapping_update: Dict[str, Type[GitService]] = None
+) -> Optional[Type[GitService]]:
+    """
+    Get the matching service class from the url.
+
+    :param url: str (url of the project, e.g. "https://github.com/packit-service/ogr")
+    :param service_mapping_update: custom mapping from  service url (str) to service class
+    :return: Matched class (subclass of GitService) or None
+    """
+    mapping = {}
+    mapping.update(_SERVICE_MAPPING)
+    if service_mapping_update:
+        mapping.update(service_mapping_update)
+
+    for service, service_kls in mapping.items():
+        if service in url:
+            return service_kls
+
+    return None
+
+
 def get_service_class(
     url: str, service_mapping_update: Dict[str, Type[GitService]] = None
 ) -> Type[GitService]:
@@ -109,12 +131,32 @@ def get_service_class(
     :param service_mapping_update: custom mapping from  service url (str) to service class
     :return: Matched class (subclass of GitService)
     """
-    mapping = {}
-    mapping.update(_SERVICE_MAPPING)
-    if service_mapping_update:
-        mapping.update(service_mapping_update)
-
-    for service, service_kls in mapping.items():
-        if service in url:
-            return service_kls
+    service_kls = get_service_class_or_none(
+        url=url, service_mapping_update=service_mapping_update
+    )
+    if service_kls:
+        return service_kls
     raise OgrException("No matching service was found.")
+
+
+def get_instances_from_dict(instances: dict) -> Set[GitService]:
+    services = []
+    for key, value in instances.items():
+        service_kls = get_service_class_or_none(url=key)
+        if not service_kls:
+            if "type" not in value:
+                raise OgrException(f"No matching service was found for url '{key}'.")
+            service_type = value["type"]
+            if service_type not in _SERVICE_MAPPING:
+                raise OgrException(
+                    f"No matching service was found for type '{service_type}'."
+                )
+
+            service_kls = _SERVICE_MAPPING[service_type]
+            value.setdefault("instance_url", key)
+            del value["type"]
+
+        service_instance = service_kls(**value)
+        services.append(service_instance)
+
+    return set(services)
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -140,12 +140,52 @@ def get_service_class(
 
 
 def get_instances_from_dict(instances: dict) -> Set[GitService]:
+    """
+    Load the service instances from the dictionary in the following form:
+
+    key = hostname, url or name that can be mapped to the service-type
+    value = dictionary with arguments used when creating a new instance of the service
+            (passed to the `__init__` method)
+
+    e.g.:
+
+    {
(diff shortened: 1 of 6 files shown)


rpitonak:
What about GitHub application? Do we want to support it also?

lachmanfrantisek:
> What about GitHub application? Do we want to support it also?

It should be supported as well -- I'll add it to the test as well.

lachmanfrantisek:
I've added tokens and other info to `__str__` methods. Is there anyone against this approach?

(I am using the `__str__` in `__hash__` implementation to allow working with sets of services.)

tests/unit/test_factory.py
```diff
@@ -213,3 +215,54 @@ def test_get_project_not_found(url, mapping, instances, exc_str):
...
+            },
+            {PagureService(token="abcd", instance_url="https://src.fedoraproject.org")},
+        ),
+        (
+            {"github.com": {"token": "abcd"}, "gitlab": {"token": "abcd"}},
+            {GithubService(token="abcd"), GitlabService(token="abcd")},
+        ),
+        (
+            {"github.com": {"github_app_id": "abcd", "github_app_private_key": "efgh"}},
+            {GithubService(github_app_id="abcd", github_app_private_key="efgh")},
```
lachmanfrantisek: This is a test for the github_app.

One downside is the need for private_key and not the path to the private_key. It is more a problem with `GithubService.__init__` method. We can probably allow both arguments in it. (We can create an issue for that and move the loading code from Packit here.)

ogr/factory.py
```diff
@@ -117,4 +117,46 @@ def get_service_class(
...
+    """
+    service_kls = get_service_class_or_none(
+        url=url, service_mapping_update=service_mapping_update
+    )
+    if service_kls:
+        return service_kls
     raise OgrException("No matching service was found.")
+
+
+def get_instances_from_dict(instances: dict) -> Set[GitService]:
```
TomasTomecek: any strong reason for the set?

TomasTomecek: also docstring please: document the format of instances

ogr/factory.py
```diff
@@ -117,4 +117,46 @@ def get_service_class(
...
+                raise OgrException(f"No matching service was found for url '{key}'.")
+            service_type = value["type"]
+            if service_type not in _SERVICE_MAPPING:
+                raise OgrException(
+                    f"No matching service was found for type '{service_type}'."
+                )
+
+            service_kls = _SERVICE_MAPPING[service_type]
+            value.setdefault("instance_url", key)
+            del value["type"]
```
TomasTomecek: I'd probably log.warn in the else branch: the user should know that the key is unknown to us

ogr/factory.py
```diff
@@ -117,4 +117,46 @@ def get_service_class(
...
+            if "type" not in value:
+                raise OgrException(f"No matching service was found for url '{key}'.")
+            service_type = value["type"]
+            if service_type not in _SERVICE_MAPPING:
+                raise OgrException(
+                    f"No matching service was found for type '{service_type}'."
+                )
+
+            service_kls = _SERVICE_MAPPING[service_type]
+            value.setdefault("instance_url", key)
```
TomasTomecek: but `key` is not an URL, it's a hostname - I know that the proposed approach will work, but let's be a bit more consistent

ogr/factory.py
```diff
@@ -117,4 +117,46 @@ def get_service_class(
...
+            del value["type"]
+
+        service_instance = service_kls(**value)
```
TomasTomecek: this totally need good docs: people have no idea what the kwargs are meant to be

tests/unit/test_factory.py
TomasTomecek: makes sense to create the issue

ogr/factory.py
lachmanfrantisek: > also docstring please: document the format of instances

Of course.

lachmanfrantisek: > any strong reason for the set?

Since the order is not importent. (Also easier testing is the reason.)

If you feel it's not necessary, I can change that to list and remove the `__hash__` implementations.

ogr/factory.py
lachmanfrantisek: Not sure what you want:
- rename of the `instance_url` => problems with deprecation
- documentation, that the key has to be hostname

ogr/factory.py
lachmanfrantisek: In the case of unknown key and specified `type`?

It is not an exceptional behaviour -- you are giving another info and you know that the mapping does not work.

ogr/factory.py
TomasTomecek: > set

I don't care to be honest, I was just curious.

ogr/factory.py
TomasTomecek: I meant this:
```python
get_instances_from_dict({
  "github.com": "token": "asd",
  "trolololo.lol": "token": "qwe"
})  
```
I'd say that user should know that trololo.lol is not in our mapping. I'd personally even raise an exc. With the current implementation, you just skip over it.

ogr/factory.py
TomasTomecek: honestly, I was literally thinking of doing "https://github.com/" instead of "github.com"

feel free to ignore if you think it's a bad idea or doesn't make sense

it does to me since it's explicit, "github.com" may even imply it's not an url

ogr/factory.py
TomasTomecek: okay, now that I look at the code, I'm confused actually.

lachmanfrantisek: Ok, the code needs some care/docs.

We are raising an exception when:
- do not find the key in mapping and there is no `type` in the `value` dictionary
    - (=we do not know the mapping)
- do not find the key in mapping, there is `type` in the `value` dictionary, but it is not in the mapping
    - (=we do not know the mapping, user specify custom mapping, but we do not know it)

TomasTomecek: got it, thanks for explanation, lgtm

ogr/factory.py
TomasTomecek: okay, let's leave it as it is now

TomasTomecek:
I'm personally still waiting for the documentation :)

lachmanfrantisek:
> I'm personally still waiting for the documentation :)

Sorry. I've been out for some time. (The docs were missing because I wanted to wait for the final design. I'll update the PR today.)

tests/unit/test_factory.py
lachmanfrantisek: Done: https://github.com/packit-service/ogr/issues/162

lachmanfrantisek: Done: https://github.com/packit-service/ogr/issues/162

lachmanfrantisek:
> I'm personally still waiting for the documentation :)

@TomasTomecek PTAL

Issue #161, opened 2019-08-29
pagure: support new way of creating PRs from forks


TomasTomecek:
https://src.stg.fedoraproject.org/api/0/#pull-requests

we just need to use the new fields such as repo_from which were added to the API in order to create PRs from forks to parent repos.

follow up to https://github.com/packit-service/packit/pull/496
blocks https://github.com/packit-service/packit.dev/pull/42

likely packit will need some code changes to support this

lachmanfrantisek:
@TomasTomecek I am not sure if need to change anything:

> Open a new pull-request from this project to itself or its parent (if this project is a fork).

Since we are using the default behaviour, it means:

- Using the `origin` instance, it will create a pull-request from `origin` to `origin`.
- Using the instance of the `fork`, it will create a pull-request from `fork` to `origin`.
- **We are missing the functionality to create pull-request from `fork` to `fork` itself and from `fork` to `origin` using the `origin` instance (=with the token from the `origin` repo?). Is that what we need?**

TomasTomecek:
> Using the instance of the fork, it will create a pull-request from fork to origin.

It won't, try it on stg. Pagure 5.7 changed this.

Reproducer: https://pagure.io/fedora-infrastructure/issue/8115#comment-591170

lachmanfrantisek:
> It won't, try it on stg. Pagure 5.7 changed this.

Do I get it right, that this is a bug on the Pagure site? So I am not sure how you want to fix that. (We can discuss that tomorrow in person.)

TomasTomecek:
yes, let's discuss in person

lachmanfrantisek:
OK, the docs are outdated and the current version works by default fork->fork when running from fork and upstream->upstream when running from upstream.

We need to use the `repo_from` fields to be able to achieve our `fork->upstream` approach. (We agreed that we want to leave the current API of OGR as is and fix only the implementation.)

PR #180, opened 2019-09-06
Fix creating Pagure PRs

lachmanfrantisek:
- Test creating Pagure PRs and fix some username problems in Pagure tests.
- Creating Pagure PRs calls upstream project url (not the url of fork).

Fixes: #161

### DIFF
--- a/tests/integration/test_pagure.py
+++ b/tests/integration/test_pagure.py
@@ -11,11 +11,13 @@ PERSISTENT_DATA_PREFIX = os.path.join(
     os.path.dirname(os.path.realpath(__file__)), DATA_DIR
 )
 
+LAST_GENERATED_BY = "lachmanfrantisek"
+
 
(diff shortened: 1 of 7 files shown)


tests/integration/test_pagure.py
```diff
@@ -11,11 +11,13 @@
     os.path.dirname(os.path.realpath(__file__)), DATA_DIR
 )
 
+LAST_GENERATED_BY = "lachmanfrantisek"
```
TomasTomecek: lol :D

lachmanfrantisek: There is a problem that when we run the tests without user-token, we need to know the owner of the fork we want to check. (=user who pregenerated the files)

I would like to improve that -- I'll create an issue.

lachmanfrantisek: https://github.com/packit-service/ogr/issues/181

PR #182, opened 2019-09-09
Support GitlabAuthenticationError

lachmanfrantisek:
- Not overwrite the gitlab token when set on read-mode.
- Support GitlabAuthenticationError in response files as well.


lachmanfrantisek:
> seems well. thanks. but at least would be nice to test the exception if this is not tested already, to check if it raises proper one

I've tested that locally. Is that enough or you want to add a special testcase for that?

lachmanfrantisek:
gate

PR #201, opened 2019-09-17
use requre for storing data for tests

jscotka:
replace storing requests objects by ``requre`` project.

- [x] regenerate test data for ogr
- [x] remove old functionality after packit will be fixed as well, not now in this PR


tests/integration/test_github.py
```diff
@@ -396,12 +396,20 @@ def test_create_release(self):
         Raise the number in `tag` when regenerating the response files.
         (The `tag` has to be unique.)
         """
-        count_before = len(self.hello_world_project.get_releases())
+        releases_before = self.hello_world_project.get_releases()
+        latest_release = releases_before[0].tag_name
```
lachmanfrantisek: This should do the trick:

```suggestion
        latest_release = self.hello_world_project.get_latest_release()
```

lachmanfrantisek: Ok, you are checking count difference, so we need to get all releases either way.

tests/integration/test_gitlab.py
```diff
@@ -143,7 +140,7 @@ def test_full_repo_name(self):
 
     def test_get_owners(self):
         owners = self.project.get_owners()
-        assert {"lachmanfrantisek", "lbarcziova"} == set(owners)
+        assert set(("lachmanfrantisek", "lbarcziova")).issubset(set(owners))
```
lachmanfrantisek:
```suggestion
        assert {"lachmanfrantisek", "lbarcziova"} <= set(owners)
```

https://docs.python.org/3.7/library/stdtypes.html#frozenset.issubset

tests/integration/test_gitlab.py
```diff
@@ -176,23 +180,33 @@ def test_issue_info(self):
...
+        issue = self.project.create_issue(title=issue_title, description=issue_desc)
+        assert issue.title == issue_title
+        assert issue.description == issue_desc
 
     def test_close_issue(self):
-        issue_for_closing = self.project.get_issue_info(issue_id=2)
+        """
+        see class comment in case of fail
+        """
+        issue = self.project.create_issue(
```
lachmanfrantisek: I don't like creating an issue in this test, but we can leave it here so the regeneration is easier.

tests/integration/test_pagure.py
```diff
@@ -297,3 +264,61 @@ def test_create_fork(self):
 
         new_forks = self.ogr_project.service.user.get_forks()
         assert len(old_forks) == len(new_forks) - 1
+
+
+class PagureProjectTokenCommands(PagureTests):
+    def setUp(self):
+        self.token = os.environ.get("PAGURE_OGR_TEST_TOKEN")
```
lachmanfrantisek: :+1:

lachmanfrantisek:
rebased and used in #231

PR #206, opened 2019-09-18
methods for commit status and commit comment

### LINKED ISSUE #172: Implement GitLab methods for commits
```python
    def commit_comment(
        self, commit: str, body: str, filename: str = None, row: int = None
    ) -> "CommitComment":
        raise NotImplementedError()

    def set_commit_status(
        self, commit: str, state: str, target_url: str, description: str, context: str
    ) -> "CommitFlag":
        raise NotImplementedError()

    def get_commit_statuses(self, commit: str) -> List[CommitFlag]:
        """
        Something like this:
        commit_object = self.gitlab_repo.commits.get(commit)
        raw_statuses = commit_object.statuses.list()
        return [
            GitlabProject._commit_status_from_gitlab_object(raw_status)
            for raw_status in raw_statuses
        ]
        """
        raise NotImplementedError()
```

- https://python-gitlab.readthedocs.io/en/stable/gl_objects/commits.html
- https://python-gitlab.readthedocs.io/en/stable/gl_objects/commits.html#commit-comments
- https://python-gitlab.readthedocs.io/en/stable/gl_objects/commits.html#commit-status

lbarcziova:
Fixes #172

### DIFF
--- a/ogr/services/gitlab.py
+++ b/ogr/services/gitlab.py
@@ -310,24 +310,74 @@ class GitlabProject(BaseGitProject):
     def commit_comment(
         self, commit: str, body: str, filename: str = None, row: int = None
     ) -> "CommitComment":
-        raise NotImplementedError()
+        """
+        Create comment on a commit.
+
+        :param commit: str The SHA of the commit needing a comment.
+        :param body: str The text of the comment
+        :param filename: str The relative path to the file that necessitates a comment
+        :param row: int Line index in the diff to comment on.
+        :return: CommitComment
+        """
+        try:
+            commit_object = self.gitlab_repo.commits.get(commit)
+        except gitlab.exceptions.GitlabGetError:
+            logger.error(f"Commit {commit} was not found.")
+            raise GitlabAPIException(f"Commit {commit} was not found.")
+
+        if filename and row:
+            raw_comment = commit_object.comments.create(
+                {"note": body, "path": filename, "line": row, "line_type": "new"}
+            )
+        else:
+            raw_comment = commit_object.comments.create({"note": body})
+        return self._commit_comment_from_gitlab_object(raw_comment, commit)
 
     def set_commit_status(
         self, commit: str, state: str, target_url: str, description: str, context: str
     ) -> "CommitFlag":
-        raise NotImplementedError()
+        """
+        Create a status on a commit
+
+        :param commit: The SHA of the commit.
+        :param state: The state of the status.
+        :param target_url: The target URL to associate with this status.
+        :param description: A short description of the status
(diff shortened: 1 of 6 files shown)


tests/integration/test_gitlab.py
```diff
@@ -97,6 +97,23 @@ def test_parent_project(self):
         assert self.project.get_fork().parent.namespace == "packit-service"
         assert self.project.get_fork().parent.repo == "ogr-tests"
 
+    def test_commit_statuses(self):
+        flags = self.project.get_commit_statuses(
+            commit="24c86d0704694f686329b2ea636c5b7522cfdc40"
+        )
+        assert isinstance(flags, list)
+        assert len(flags) == 0
```
lachmanfrantisek: It would be nice to have also some tests where we have some flags.

You can rename this test to something like `test_commit_statuses_empty` and add one with some flags...

ogr/services/gitlab.py
```diff
@@ -310,24 +310,74 @@ def pr_create(
...
+        else:
+            raw_comment = commit_object.comments.create({"note": body})
+        return self._commit_comment_from_gitlab_object(raw_comment, commit)
 
     def set_commit_status(
         self, commit: str, state: str, target_url: str, description: str, context: str
     ) -> "CommitFlag":
-        raise NotImplementedError()
+        """
+        Create a status on a commit
```
lachmanfrantisek: Could you please add a test for this?

lbarcziova:
Thanks, I will add that tests. I tried to regenerate the failing test with the new version of python-gitlab, but it didn't work.

lachmanfrantisek:
> Thanks, I will add that tests. I tried to regenerate the failing test with the new version of python-gitlab, but it didn't work.

I've been able to regenerate `test_commit_comment`. Would you like me to push that file?

The `test_commit_statuses` then fail for me, but it is probably due to the empty list that is returned.

lbarcziova:
It's not neccessary, `test_commit_comment` worked for me too. I have just changed the tests and it seems that it works now.

tests/integration/test_gitlab.py
```diff
@@ -97,12 +97,29 @@ def test_parent_project(self):
...
 
-    def test_commit_statuses(self):
+    def test_get_commit_statuses(self):
         flags = self.project.get_commit_statuses(
-            commit="24c86d0704694f686329b2ea636c5b7522cfdc40"
+            commit="11b37d913374b14f8519d16c2a2cca3ebc14ac64"
         )
         assert isinstance(flags, list)
-        assert len(flags) == 0
+        assert len(flags) >= 2
```
lachmanfrantisek: Nice, tests passed!

Could you also check the content?

PR #217, opened 2019-09-22
Implement GitProject.get_web_url()

### LINKED ISSUE #213: Implement GitProject.get_url()
In GitProject we do not have a method for getting web URL of the project.

AC:

- [ ] Add the method to `abstract.GitProject`. (And raise `NotImplementedError`.)
- [ ] Implement the method for all forges (GitHub, GitLab, Pagure).
- [ ] Add tests for this method.

Links:
- https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html
- https://python-gitlab.readthedocs.io/en/stable/gl_objects/projects.html
    https://docs.gitlab.com/ce/api/projects.html
- https://pagure.io/api/0/#projects

Examples:
- https://github.com/packit-service/ogr
- https://gitlab.com/packit-service/ogr-tests
- https://pagure.io/ogr-tests/

mfocko:
Closes #213
Seems like it could be implemented in BaseGitProject.
Not quite sure about Pagure implementation, since I haven't found any tests regarding `None` as namespace.


mfocko:
I cannot seem to find any way to get URL for Pagure project in API.
Also I think that it would be better to remove condition from the function `get_web_url` by fixing non-existing namespace either in BaseGitProject or override the property in PagureProject

lachmanfrantisek:
> I cannot seem to find any way to get URL for Pagure project in API.

Maybe we can use  `url_path` property from projet's info (`PagureProject.get_project_info`):

- https://src.fedoraproject.org/api/0/rpms/python-docker (with namespace)
- https://pagure.io/api/0/ogr-tests (without namespace)

It looks like it takes care of the namespace.

> Also I think that it would be better to remove handling no-namespace from the function get_web_url by fixing non-existing namespace either in BaseGitProject or override the property in PagureProject

I am not sure I fully get your point but feel free to send that change, and we can discuss the proposal here.

mfocko:
> Maybe we can use `url_path` property from projet's info (`PagureProject.get_project_info`):
> ...
> It looks like it takes care of the namespace.

Updated, but it still uses `instance_url` from service.

> I am not sure I fully get your point but feel free to send that change, and we can discuss the proposal here.
Yeah, sorry, I didn't mention what I'm talking about, `full_repo_name` property in `BaseGitProject`...
1. Overriding it in `PagureProject` should do
1. I've just realised it's implemented twice `BaseGitProject` and `GitProject`

lachmanfrantisek:
> Fix flake8 remarks

Nice! I forgot to mention [pre-commit](https://github.com/packit-service/ogr/blob/master/CONTRIBUTING.md#checkerslintersformatters--pre-commit) yesterday.

lachmanfrantisek:
@mfocko Sorry, I missed your previous comment before merging...

> Updated, but it still uses instance_url from service, since url_path doesn't contain it.

That's fine.

> Yeah, sorry, I didn't mention what I'm talking about, full_repo_name property in BaseGitProject...
>  - Overriding it in PagureProject should do
>  - I've just realised it's implemented twice (in BaseGitProject and GitProject)

You are right. If you want, create a new pull request with that:
- Remove the code from `GitProject` -- it should be abstract (`raise NotImplementedError()`)
- Override that in `PagureProject` using the `url_path`

PR #221, opened 2019-09-25
Remove pull requests from issues in GitHub implementation

### LINKED ISSUE #107: Remove PullRequests from list of Github Issues.
Based on [Github API doc](https://developer.github.com/v3/pulls/#labels-assignees-and-milestones):
>Every pull request is an issue, but not every issue is a pull request.

Therefore, in the case of Github, functions as `get_issue_list`, `get_issue_info` etc. are working also with Pull Requests and behavior isn't tested.

We should get rid of Pull Requests in functions working with Issues. Pagure doesn't have this problem. 

--------------
@lachmanfrantisek added:

To be able to be compatible with other git forges, let's differentiate between issues and pull-requests. (And do not allow working with GitHub issues, that are actually pull-requests. Issues and pull-requests are distinct sets.)

AC:

- [ ] Go through the issue related code in `GithubProject` and check if it does not return any PullRequest.
- [ ] Create test cases for that.

mfocko:
Closes #107


tests/integration/test_github.py
```diff
@@ -283,6 +283,26 @@ def test_issue_labels(self):
...
 
+    def test_list_contains_only_issues(self):
+        issue_list_all = self.ogr_project.get_issue_list(status=IssueStatus.all)
+        issue_ids = [issue.id for issue in issue_list_all]
+
+        pr_ids = [219, 207, 201, 217, 208, 210]
+        for id in pr_ids:
+            assert id not in issue_ids
+
+    def test_get_issue(self):
```
lachmanfrantisek: The name of the test does not fully express what the test does.

PR #226, opened 2019-09-26
Make PagureProject.full_repo_name property offline

### LINKED ISSUE #220: PagureProject.full_repo_name touches service
This property violates our rule that only methods with `get` can touch the real servers.

We can:

- Revert the change and stay with guessing.
- Change the property to `get_full_repo_name` (needs deprecation to original property)
- Change our mind and do not be strict about that. (And use properties instead of methods for objects.)
    - This will bring nicer user-experience, but maybe some confusion.
    - Maybe, we can leave only `__init__` offline?

mfocko:
Fixes #220


mfocko:
So should I wat for the release?
I could also add test of forked repo with namespace... Seems to keep all parts of URL, it's handled

lachmanfrantisek:
> So should I wat for the release?

No, just note that I would like to have this in the `0.8.0` release..;)

> I could also add test of forked repo with namespace... Seems to keep all parts of URL, it's handled

That would be nice. Thanks.

mfocko:
1. `PagureProject._user` queries `whoami`

lachmanfrantisek:
>     1. `PagureProject._user` queries `whoami`

I don't think so. We are using the value from the `__init__`:

```python
class PagureProject(BaseGitProject):
    service: PagureService

    def __init__(
        self,
        repo: str,
        namespace: Optional[str],
        service: "PagureService",
        username: str = None,
        is_fork: bool = False,
    ) -> None:
        super().__init__(repo, service, namespace)
        self.read_only = service.read_only

        self._is_fork = is_fork
        self._username = username

        self.repo = repo
        self.namespace = namespace

    @property
    def _user(self) -> str:
        if not self._username:
            self._username = self.service.user.get_username()
        return self._username
```

But you are right that this property is not offline as well.


The response files contain the project description. The problem is here:

```python
    @property
    def is_fork(self) -> bool:
        return bool(self.get_project_info()["parent"])
 ```

Another property that is not offline..;(

We have attribute `_is_fork` which is offline and is set in `__init__` and we have a property `is_fork` that is not offline and uses `get_project_info`.

-----------

Do you want to fix all of them?

- I am not sure if we can use `self._username` instead of `self._user` without problems. (So I am not sure if we can do `full_repo_name` offline..;)
- We can probably use `_is_fork` in all places. (Or remove `is_fork` property and rename `_is_fork` to `is_fork`.)

What do you think? I hope I don't miss any problem that can be caused by this.

**Comment by lachmanfrantisek** [comment]
It looks harder now than it seemed at the start.

**Comment by mfocko** [comment]
also those tests for project with namespaces call `PagureService.get_project`

**Comment by lachmanfrantisek** [comment]
We need to clean the logic with the attributes in `__init__` and the properties. (It is a mess in the current state.)

To make it clear what we need to do now:

- If it is possible, implement the `full_repo_name` offline, please.
- If it is not possible, we need to setup deprecation policy (on of your main tasks..;) and move from properties to methods.

Is that ok?

**Comment by mfocko** [comment]
`full_repo_name` is basically offline, unless the username is not defined...
ok, i'm confused now...
```python
assert self.ogr_project
assert self.service.get_project(namespace="mbi", repo="ansible")
assert self.ogr_fork
assert self.service.get_project(
  namespace="Fedora-Infra",
  repo="ansible",
  username=self.user,
  is_fork=True
)
```
This only queries `whoami`... Once I add `full_repo_name` it queries project info... from which `namespace`, `repo` is defined... only `_user` is not defined... but why it queries whole project info?

lachmanfrantisek:
>  basically offline, unless the username is not defined...

This is not offline to be correct..;)

> Once I add full_repo_name it queries project info... from which namespace, repo is defined... only _user is not defined... but why it queries whole project info?

Because of the usage of this:

```python
    @property
    def is_fork(self) -> bool:
        return bool(self.get_project_info()["parent"])
```

mfocko:
> This is not offline to be correct..;)

i know :D

> Because of the usage of this:
> ...

Probability of me putting the `full_repo_name` randomly on fork, 100%

mfocko:
- [ ] make `PagureProject.is_fork` offline (checked in `full_repo_name` because of prefix)
 btw why it checks against `project_info`?

- [ ] `PagureProject._user`
 pass it to constructor when getting project from `PagureService`?

lachmanfrantisek:
> make PagureProject.is_fork offline (checked in full_repo_name because of prefix)
btw why it checks against project_info?

We are checking `parent` key.

> pass it to constructor when getting project from PagureService?

We can get the username from API in `PagureService.get_project` to not force users to specify that since we can get it from the token. And we can touch the API in methods..;)

mfocko:
> We are checking `parent` key.

I've seen it, but why are we passing the `is_fork` in constructor then if it's "ignored"? :D
(Used in two places... `__eq__` and `_get_project_url`)

> We can get the username from API in `PagureService.get_project` to not force users to specify that since we can get it from the token. And we can touch the API in methods..;)

That's what I meant, just left the context out...

lachmanfrantisek:
> I've seen it, but why are we passing the `is_fork` in constructor then if it's "ignored"? :D
> (Used in two places... `__eq__` and `_get_project_url`)

The `_get_project_url` is internally used in most of the calls.

> That's what I meant, just left the context out...

:+1:

mfocko:
Just to make sure if I understand correctly, `is_fork` checks whether the project is **really** fork, since it queries API?
Also, if I pass `is_fork=False` to the constructor and the property returns `True`, what's expected behaviour? Throwing exception or correcting `_is_fork`?

That means I could use `_is_fork` in `full_repo_name`, that wouldn't call API.

`PagureProject.parent` (property) also uses `is_fork`

lachmanfrantisek:
> Just to make sure if I understand correctly, is_fork checks whether the project is really fork, since it queries API?

:+1: 

> Also, if I pass is_fork=False to the constructor and the property returns True, what's expected behaviour? Throwing exception or correcting _is_fork?

I think this should not occur for valid input. (Maybe only if you try something like `get_project(is_fork=False, repo="python-ogr", namespace="fork/lachmanfrantisek/rpms")`. The namespace shoult be only `rpms`.) 

> That means I could use _is_fork in full_repo_name, that wouldn't call API.

Yes, we can probably remove `is_fork` property completely. (Or implement it via `self._is_fork`.)

> PagureProject.parent (property) also uses is_fork

This is probably not doable offline. (The parent can have different name and we cannot guess that.)

mfocko:
If we need it in `parent`, what about swapping them?
Make the `is_fork` that uses API private (I'm not sure if it can be private property) and expose the one given in constructor

lachmanfrantisek:
> If we need it in `parent`, what about swapping them?
> Make the `is_fork` that uses API private (I'm not sure if it can be private property) and expose the one given in constructor

Yes, the dynamic one can be method (not a property). What about renaming it to something like `get_is_fork_from_api` to avoid confusion?

mfocko:
Fine with me

mfocko:
passing username to project when getting it from service creates more `whoami` queries

lachmanfrantisek:
> passing username to project constructor when getting it from service creates more `whoami` queries

It is not a problem if the calls are done outside of the `__init__` or `full_repo_name`.

lachmanfrantisek:
Do you have some problems with the second part?

> - [ ] PagureProject._user
> pass it to constructor when getting project from PagureService?

mfocko:
> Do you have some problems with the second part?
> 
> > * [ ]  PagureProject._user
> >   pass it to constructor when getting project from PagureService?

not yet, but will do :D conflicts with get_issue_comments tests

lachmanfrantisek:
> not yet, but will do :D conflicts with get_issue_comments tests

OK...;)

ogr/services/pagure.py
```diff
@@ -127,7 +127,12 @@ def __hash__(self) -> int:
         return hash(str(self))
 
     def get_project(self, **kwargs) -> "PagureProject":
-        return PagureProject(service=self, **kwargs)
+        if "username" in kwargs:
+            return PagureProject(service=self, **kwargs)
+        else:
+            return PagureProject(
+                service=self, username=self.user.get_username(), **kwargs
+            )
```
lachmanfrantisek: What do you think about this solution?

(Just a suggestion. Pick the one you like more.)

```python
    def get_project(self, username=None, **kwargs) -> "PagureProject":
        username = username or self.user.get_username()
        return PagureProject(
            service=self, username=self.user.get_username(), **kwargs
        )
```
----------

Btw, these two code snippets are equivalent..;-)

```python
if something:
    return "a"
else:
    return "b"
```
```python
if something:
    return "a"
return "b"
```

mfocko: Yeah, I know...
Why does it remind me of JS...

lachmanfrantisek: Yes..;)  However, those oneliners usually bring no readability benefit.

mfocko: also that rewrite broke the tests...

lachmanfrantisek: I do not see the difference.

mfocko: thought it's side effect of `or` evaluating expression before checking first one... tested in REPL, should not be

lachmanfrantisek:
```python
In [1]: def fce(): 
   ...:     print("hello") 
   ...:     return "a" 
   ...:                                                                                                                                                                                                                                        

In [2]: a = "a" or fce()                                                                                                                                                                                                                       

In [3]: a = None or fce()                                                                                                                                                                                                                      
hello
```

Sorry, but I don't understand.

Only difference I can see is when in `kwargs` you have `'username': None`.

mfocko: your suggestion breaks 4 tests

```python
username = kwargs["username"] if "username" in kwargs else self.user.get_username()
return PagureProject(
  service=self, username=username, **kwargs
)
```
this breaks 10...

mfocko: Also what I've noticed, there are some intentionally failing tests for `get_project` like getting github project out of it... I don't understand why they were fine with `if ... else ...` but not with that

lachmanfrantisek: My fault, I had an error in my code -- I've forgotten to use the username variable in the PagureProject call. (I've fixed that in the comment.)

mfocko: I was using the username

lachmanfrantisek: Hmm, that's weird.

> atter one has the problem with multiple values for 'username'

How is that possible?

I'll try to play with your code locally to see what is happening there.

lachmanfrantisek: In the meantime, go with the solution that works better for you.

mfocko: > How is that possible?

this one (the one that broke the record of broken tests):
```python
username = kwargs["username"] if "username" in kwargs else self.user.get_username()
return PagureProject(
  service=self, username=username, **kwargs
)
```
it does pass it two times, once via username and second time with `**`

mfocko: this works:
```python
def get_project(self, username=None, **kwargs) -> "PagureProject":
    if username:
        return PagureProject(
            service=self, username=username, **kwargs
        )

    return PagureProject(
        service=self, **kwargs
    )
```
and this doesn't:
```python
def get_project(self, username=None, **kwargs) -> "PagureProject":
    if username:
        return PagureProject(
            service=self, username=username, **kwargs
        )

    return PagureProject(
        service=self, username=self.user.get_username(), **kwargs
    )
```

Also it appears to cause problems with "invalid" tests:
1. trying to get project from URL with `github.com` repository
1. `pagure.io/ogr` -> non-existing
1. `src.fedoraproject.org` didn't make token for that, so that could be reason of failing on my side

lachmanfrantisek: > Also it appears to cause problems with "invalid" tests:

It started to touch the remote services what was not expected when writing the code. (The test uses mapping like `{"github.com": PagureService}` what is fine until we try to call an API.)

Let's use the current approach and spend some time on `offline x online` handling later.

lachmanfrantisek:
> I'll create an issue for the future investigation

Created: #229

lachmanfrantisek:
Because of the change in response files, I need to release a new version before merging this to not break the projects on top of ogr. (We will change response files in #201 so it won't be a problem after release.)

lachmanfrantisek:
@mfocko The new version was release -- can you rebase this PR, please? Thanks!

lachmanfrantisek:
recheck

PR #227, opened 2019-09-27
Factor out getting collaborators (GithubProject)

### LINKED ISSUE #224: Identical implementation for collaborators (Github Project)
In `GithubProject` class there is identical implementation for `who_can_close_issue` and `who_can_merge_pr`, is it intentional? Or can we factor it out since it's identical?

mfocko:
Closes #224

### DIFF
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -350,7 +350,7 @@ class GithubProject(BaseGitProject):
         # in case of github, repository has only one owner
         return [self.github_repo.owner.login]
 
-    def __collaborators(self) -> Set[str]:
+    def __get_collaborators(self) -> Set[str]:
         try:
             collaborators = self._get_collaborators_with_permission()
         except github.GithubException:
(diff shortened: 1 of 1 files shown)


ogr/services/github.py
```diff
@@ -366,21 +366,11 @@ def who_can_close_issue(self) -> Set[str]:
...
-                f"Current Github token must have push access to view repository permissions."
-            )
-            return set()
-
-        usernames = []
-        for login, permission in collaborators.items():
-            if permission in ["admin", "write"]:
-                usernames.append(login)
+    def who_can_close_issue(self) -> Set[str]:
+        return self.__collaborators()
```
lachmanfrantisek: The name can be a little bit confusing -- I think that no all collaborators have that permission. (Also you can use some verb in the name like `get`.)

mfocko: it's given by interface... another refactor?

lachmanfrantisek: I mean the `__collaborators`.

mfocko: oh ok :D

PR #228, opened 2019-09-27
Implementation of get_issue_comments for projects

### LINKED ISSUE #205: Fix/implement get_issue_comments
In abstract class `GitProject`, we have the following method:

```python
    def get_issue_comments(
        self, issue_id, filter_regex: str = None, reverse: bool = False
    ) -> List["IssueComment"]:
        """
        Get list of Issue comments.

        :param issue_id: int
        :param filter_regex: filter the comments' content with re.search
        :param reverse: reverse order of comments
        :return: [IssueComment]
        """
        raise NotImplementedError()
```

This issue is not implemented in `GithubProject`, but the real implementation should be in the `_get_all_issue_comments`.

```python
    def _get_all_issue_comments(self, issue_id: int) -> List[IssueComment]:
        issue = self.github_repo.get_pull(number=issue_id)
        return [
            self._issuecomment_from_github_object(raw_comment)
            for raw_comment in issue.get_issue_comments()
        ]
```

AC:
- [ ] Github/Pagure
    - [ ] implement the `get_issue_comments` with the code `_get_all_issue_comments` (There are some missing parameters.)
    - [ ] remove the `_get_all_issue_comments` and change its usage to `get_issue_comments`
    - [ ] make sure it works as expected
- [ ] Implement the `get_issue_comments` in `GitlabProject`
- [ ] create test(s) for that if they are not already there

mfocko:
Closes #205

- [x] `GithubProject.get_issue_comments`

- [x] `GitlabProject.get_issue_comments`

- [x] `PagureProject.get_issue_comments`

### DIFF
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -426,7 +426,9 @@ class GithubProject(BaseGitProject):
         issue = self.__get_issue(number=issue_id)
         return self._issue_from_github_object(issue)
 
-    def get_issue_comments(self, issue_id: int, filter_regex: str = None, reverse: bool = False) -> List[IssueComment]:
+    def get_issue_comments(
+        self, issue_id: int, filter_regex: str = None, reverse: bool = False
+    ) -> List[IssueComment]:
         """
         Get list of Issue comments.
 
--- a/ogr/services/github.py
+++ b/ogr/services/github.py
@@ -22,6 +22,7 @@
 
 import logging
 from pathlib import Path
+import re
 from typing import Optional, Dict, List, Type, Set
 
 import github
@@ -425,12 +426,29 @@ class GithubProject(BaseGitProject):
         issue = self.__get_issue(number=issue_id)
         return self._issue_from_github_object(issue)
 
-    def _get_all_issue_comments(self, issue_id: int) -> List[IssueComment]:
+    def get_issue_comments(self, issue_id: int, filter_regex: str = None, reverse: bool = False) -> List[IssueComment]:
+        """
+        Get list of Issue comments.
+
+        :param issue_id: int
+        :param filter_regex: filter the comments' content with re.search
+        :param reverse: reverse order of comments
+        :return: [IssueComment]
+        """
         issue = self.__get_issue(number=issue_id)
-        return [
+        if not filter_regex:
+            filter_regex = r".*"
+
+        comments = [
             self._issuecomment_from_github_object(raw_comment)
-            for raw_comment in issue.get_issue_comments()
+            for raw_comment in issue.get_comments()
+            if re.search(filter_regex, raw_comment.body)
         ]
+
+        if reverse:
+            comments.reverse()
+
+        return comments
 
     def issue_comment(self, issue_id: int, body: str) -> IssueComment:
         """
--- a/ogr/services/gitlab.py
+++ b/ogr/services/gitlab.py
@@ -21,6 +21,7 @@
 # SOFTWARE.
 
 import logging
+import re
 from typing import List, Optional, Dict, Set
 
 import gitlab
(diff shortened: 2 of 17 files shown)


mfocko:
**blocked**
seemes like a Pagure API bug, can't retrieve Issue with token (returns invalid/expired token, reason: unauthorized)
without token there's no problem

lachmanfrantisek:
> **blocked**
> seems like a Pagure API bug, can't retrieve Issue with token (returns error: invalid/expired token, reason: unauthorized)
> without token there's no problem

You can generate response files without a token or generate a per-project token. (I've added you the admin rights to the Pagure testing repository.)

mfocko:
thinking about a way to refactor it... however

1. GitLab returns comments in reverse order -> _reverse_
1. Pagure doesn't have structure similar to Github and Gitlab -> _regex filtering_

So...
1. move `get_issue_comments` to base class AND
1. add `_issuecomment_from_object` to interface so it could be used in base class
1. reversed order in Gitlab could be probably fixed with returning comments wrapped in `reverse` iterator (didn't work with Github, so not sure if it would work with Gitlab)
1. regex filtering...
    1. function to retrieve the comment's body?
    1. filter them afterwards?
        _one more iteration through the comments and would also create objects that are not needed_

lachmanfrantisek:
> regex filtering... 

Get inspired by the `get_pr_comment`.

It would be nice to handle that in the same way:
- If you find some common code, use some common helper-function.
- If you find some problem, let's fix that for PRs as well for issues.

mfocko:
The `IssueComment` and `PRComment` are same... Can't we extract it to `Comment`?
Probably make `IssueComment` and `PRComment` subclasses?

lachmanfrantisek:
> The `IssueComment` and `PRComment` are same... Can't we extract it to `Comment`?

Currently yes, but I would like to add a link to the original PR/Issue when working on #86 .

> Probably make `IssueComment` and `PRComment` subclasses?

:+1: Makes sense.

What do you think about this structure?

- *Comment*
    - *IssueComment*
        - GithubIssueComment
        - PagureIssueComment
        - GitlabIssueComment
    - *PRComment*
        - GithubPRComment
        - PagurePRComment
        - GitlabPRComment

*Abstract classes are in italics.* Maybe we can also add a middle peace (e.g. `BaseComment` or `BaseIssueComment`) if needed as we do with projects.

mfocko:
So what about merging this and creating separate issue related to #86?
Since it involves more than just implementation of `get_issue_comments`

lachmanfrantisek:
> So what about merging this and creating separate issue related to #86?

:+1: We only need to wait with merging to #223 (Sorry, but we need to be sure what we are releasing.)

Do you want to create the new issue or describe that in that original one?

mfocko:
> Do you want to create the new issue or describe that in that original one?

Will do

lachmanfrantisek:
recheck

ogr/services/github.py
```diff
@@ -425,13 +426,32 @@ def get_issue_info(self, issue_id: int) -> Issue:
...
             self._issuecomment_from_github_object(raw_comment)
-            for raw_comment in issue.get_issue_comments()
+            for raw_comment in issue.get_comments()
+            if re.search(filter_regex, raw_comment.body)
         ]
 
+        if reverse:
+            comments.reverse()
+
+        return comments
```
lachmanfrantisek: Since the code is nearly the same for all implementation, what about:

- Implement skeleton implementation in the base class and use e.g. `_get_all_issue_comments` for getting all issue comments. (As we do for pull-request.) This can lead to spaghetti code.
- Add some filtering/sorting-helper method in the base class (or in some other place) and use them in all implementations.

mfocko: I planned to do that in #230, but ok

lachmanfrantisek: Ok, so no problem.

lachmanfrantisek: Thanks for fixing that here.

mfocko: So, revert it? And start working on #230?

lachmanfrantisek: Do it as you wish. (In the current state, we only need to make the pre-commit happy.)

mfocko: Ok, I'll revert and push...
Probably stash what I have for #230...

lachmanfrantisek: OK, choose the way you want.

PR #238, opened 2019-10-09
remove all stuff what were moved to requre project

jscotka:


TomasTomecek:
We should also remove this line:
```
os.environ["RECORD_REQUESTS"] = "TRUE"
```
from `packit/tests/testsuite_basic/integration/__init__.py`

jscotka:
> We should also remove this line:
> 
> ```
> os.environ["RECORD_REQUESTS"] = "TRUE"
> ```
> 
> from `packit/tests/testsuite_basic/integration/__init__.py`

@TomasTomecek sure. this is old relic in ``packit`` project  without no effect on ``ogr`` side.

PR #242, opened 2019-10-11
Add project_create to PagureService and add tests for it

### LINKED ISSUE #204: Implement Pagure method for create_project
```python
    def project_create(self, repo: str, namespace: str = None) -> "PagureProject":
        pass
```

API documentation:
https://pagure.io/api/0/#projects

AC:

- [ ] implement the method in `PagureService`
- [ ] create at least two tests for that (with and without specifying `namespace`)

mfocko:
Closes #204

### DIFF
--- a/ogr/services/pagure/service.py
+++ b/ogr/services/pagure/service.py
@@ -25,7 +25,7 @@ from typing import List, Optional
 
 import requests
 
-from ogr.exceptions import PagureAPIException
+from ogr.exceptions import PagureAPIException, OgrException
 from ogr.factory import use_for_service
 from ogr.parsing import parse_git_repo
 from ogr.services.base import BaseGitService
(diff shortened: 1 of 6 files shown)


ogr/services/pagure/service.py
```diff
@@ -244,5 +245,15 @@ def project_create(
...
-        self.call_api(request_url, "POST", data=parameters)
+        try:
+            self.call_api(request_url, "POST", data=parameters)
+        except PagureAPIException as ex:
+            if (
+                ex.pagure_response
+                and ex.pagure_response["errors"]["namespace"][0] == "Not a valid choice"
+            ):
+                raise OgrException(
+                    f"Cannot create project in given namespace ({namespace})."
```
lachmanfrantisek: Can we determine these two possibilities?

- non-existing namespace
- no permissions

mfocko: I remember it returning the same thing. _not a valid choice_

mfocko: yeah, returns the same thing

lachmanfrantisek: Hmm. I'll create an issue for that, so we don't forget about that.

lachmanfrantisek: #245

PR #249, opened 2019-10-16
Factor out Comment

### LINKED ISSUE #230: Factor out Comment
Part of #86

Structure of classes:

* _Comment_
  * _IssueComment_
    * GithubIssueComment
    * GitlabIssueComment
    * PagureIssueComment
  * _PRComment_
    * GithubPRComment
    * GitlabPRComment
    * PagurePRComment

suggested by @lachmanfrantisek in #228

* [x] Create a `Comment` abstract class
* [x] Implement base class for `Comment`
* [x] Implementation of services
  * [x] Github
  * [x] Gitlab
  * [x] Pagure
* [x] Update interface to include function for returning `Comment` object
  * [x] Deprecate functions (`*Project`) manipulating with comments (#121)
* [ ] Remove `Comment = TypeVar('Comment', IssueComment, PRComment)`
* [ ] Make service-specific  `__str__` to be compatible with parent's constructor
* [ ] Decide return types for services `PRComment` vs `GithubPRComment`

Maybe the implementation for specific service could take the `raw_comment` in constructor?

mfocko:
Closes #230

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -128,9 +128,9 @@ class Comment:
         created: Optional[datetime.datetime] = None,
         edited: Optional[datetime.datetime] = None,
     ) -> None:
-        if raw_comment is not None:
+        if raw_comment:
             self._from_raw_comment(raw_comment)
-        elif comment is not None and author is not None:
+        elif comment and author:
             self.comment = comment
             self.author = author
             self.created = created


ogr/abstract.py
```diff
@@ -157,14 +133,24 @@ def __init__(
...
             f"comment='{comment}', "
             f"author='{self.author}', "
             f"created='{self.created}', "
             f"edited='{self.edited}')"
         )
 
 
+class IssueComment(Comment):
+    def __str__(self) -> str:
+        return "Issue" + super().__str__()
```
lachmanfrantisek: What about the backward link to the original issue object?

ogr/abstract.py
```diff
@@ -157,14 +133,24 @@ def __init__(
...
+
+
+class PRComment(Comment):
+    def __str__(self) -> str:
+        return "PR" + super().__str__()
```
lachmanfrantisek: What about the backward link to the original pull-request object?

ogr/services/github/comments.py
```diff
@@ -0,0 +1,51 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from github import IssueComment as _GithubIssueComment
+
+from ogr.abstract import IssueComment, PRComment
+
+
+class GithubIssueComment(IssueComment):
+    def __init__(self, raw_comment: _GithubIssueComment) -> None:
```
lachmanfrantisek: I would save this object as a "private" property. What do you think?

(So we can update the comments in the future. -- e.g., I would like to have some method for hiding comments)

ogr/services/gitlab/comments.py
```diff
@@ -0,0 +1,50 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+
+from ogr.abstract import IssueComment, PRComment
+
+
+class GitlabIssueComment(IssueComment):
+    def __init__(self, raw_comment) -> None:
```
lachmanfrantisek: Same as for Github.

ogr/abstract.py
mfocko: You mean the object from package? Or the ones that are supposed to be implemented (ogr's Issue, PR)?

ogr/services/github/comments.py
mfocko: You mean not exporting it and keeping only the Issue/PR Comment exposed?

ogr/abstract.py
lachmanfrantisek: The OGR object. i would go with this logic:

```python
pr = project.pr_create(...)
assert pr.project == project
pr_comment = pr.comment(...)
assert pr_comment.pull_request == pr
```

mfocko: Not sure if it's possible right now, since we're using only the objects from github/gitlab packages, pagure doesn't even touch PR in any way (except the comments)... You want to do it in this PR or later when implementing Issue/PR?

lachmanfrantisek: > Not sure if it's possible right now

Yes, I am not sure as well. This is the ideal state I would like to achieve..;-)

Leave it for the next PR, since we would need to construct the `caller` after the real call.

ogr/services/github/comments.py
lachmanfrantisek: I mean `raw_comment` will be useful.
```suggestion
    def __init__(self, raw_comment: _GithubIssueComment) -> None:
        self.__raw_comment = raw_comment
```

lachmanfrantisek: So we can implement some methods using this `__raw_comment`.

mfocko: ok

mfocko:
* [ ] Remove `Comment = TypeVar('Comment', IssueComment, PRComment)`
* [ ] Make service-specific  `__str__` to be compatible with parent's constructor
  not sure if we expect serialization of `str` (since it's used when `repr` is not defined) or just pretty-printing
  right now, they're not compatible since only `(Issue/PR)Comment` have constructor of the same signature
* [ ] Decide return types for services `PRComment` vs `GithubPRComment`
  so are we fine returning PRComment type? or we want to return the service-specific one?
* [ ] Add TODO in `comments.py` to keep Issue/PR object when factored to separate classes

ogr/abstract.py
mfocko: added TODO in services' `comments.py`

lachmanfrantisek:
>     * Decide return types for services `PRComment` vs `GithubPRComment`
>       so are we fine returning PRComment type? or we want to return the service-specific one?

I would go with service-specific one -- this is what it will actually return. (If we use the abstract one, we will have type-problems in the implementations (non-existing `__raw_comment` property.)

lachmanfrantisek:
>     * Make service-specific  `__str__` to be compatible with parent's constructor
>       not sure if we expect serialization of `str` (since it's used when `repr` is not defined) or just pretty-printing
>       right now, they're not compatible since only `(Issue/PR)Comment` have constructor of the same signature

I like the constructor-like syntax, but showing only the `raw_comment` is not what I would like to see. Using the abstract constructor style is a good idea. (e.g. `GithubPRComment(comment='Bla bla.', author='me')`.

ogr/services/github/comments.py
```diff
@@ -29,28 +29,34 @@
...
+        if raw_comment is not None:
+            self.__raw_comment = raw_comment
+            super().__init__(
+                comment=raw_comment.body,
+                author=raw_comment.user.login,
+                created=raw_comment.created_at,
+                edited=raw_comment.updated_at,
+            )
+        else:
+            super().__init__(**kwargs)
```
lachmanfrantisek: Hmm, interesting. (Compatible with everything..;-)

Personally, I don't like that we don't see the real attributes in the `__init__`. (Only the `**kwargs**`.)

Is there a situation, where we want to use it? I am just worried about the situation, where we have two kinds of this object that needs to be handled differently.

lachmanfrantisek: Can you at least explicitly name the arguments? I know it's not necessary, but for the documentation purposes.

You know. *"Explicit is better than implicit."*

lachmanfrantisek:
recheck

ogr/abstract.py
```diff
@@ -141,29 +119,49 @@ def __str__(self) -> str:
...
+        comment: Optional[str] = None,
+        author: Optional[str] = None,
         created: Optional[datetime.datetime] = None,
         edited: Optional[datetime.datetime] = None,
     ) -> None:
-        self.comment = comment
-        self.author = author
-        self.created = created
-        self.edited = edited
+        if raw_comment is not None:
```
lachmanfrantisek:
```suggestion
        if raw_comment:
```

ogr/abstract.py
```diff
@@ -141,29 +119,49 @@ def __str__(self) -> str:
...
+            self._from_raw_comment(raw_comment)
+        elif comment is not None and author is not None:
```
lachmanfrantisek:
```suggestion
        elif comment and author:
```

Issue #251, opened 2019-10-22
Strict mypy


lachmanfrantisek:
There are many warnings and errors when running `mypy` in a strict mode. Each error can mean potentially problematic part of the code. (The goal is to make the code better, not only to silence the mypy.) 

It would be nice to reduce the number of errors/warnings.

What you can do:
- Run the mypy on the code:
  ```sh
  mypy ogr --strict
  ```
- pick an error/warning
- try to fix the problem which can mean
  - adding a type annotation
  - changing the type annotation
  - adding some test for the input and raising a propper exception
  - something else
- send a PR with the fix or raise an issue if you found some problem, that cannot be easily fixed

This does not need to be done in one PR. Also, one fixed error can be useful to us.

<details>
<summary>'mypy ogr --strict' output</summary>

```
$ mypy ogr --strict
ogr/deprecation.py:23: error: Cannot find implementation or library stub for module named 'deprecated'
ogr/deprecation.py:26: error: Function is missing a type annotation
ogr/services/github/auth_providers/abstract.py:6: error: Cannot find implementation or library stub for module named 'github'
ogr/services/github/auth_providers/abstract.py:6: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
ogr/services/github/auth_providers/abstract.py:32: error: Function is missing a type annotation for one or more arguments
ogr/parsing.py:99: error: Function is missing a return type annotation
ogr/parsing.py:155: error: Argument 1 to "RepoUrl" has incompatible type "None"; expected "str"
ogr/services/github/auth_providers/token.py:6: error: Cannot find implementation or library stub for module named 'github'
ogr/services/github/auth_providers/token.py:12: error: Function is missing a type annotation for one or more arguments
ogr/services/github/auth_providers/token.py:35: error: Function is missing a type annotation for one or more arguments
ogr/services/github/auth_providers/token.py:35: error: Incompatible default for argument "token" (default has type "None", argument has type "str")
ogr/services/github/auth_providers/github_app.py:7: error: Cannot find implementation or library stub for module named 'github'
ogr/services/github/auth_providers/github_app.py:26: error: Returning Any from function declared to return "bool"
ogr/services/github/auth_providers/github_app.py:61: error: Incompatible return value type (got "None", expected "str")
ogr/services/github/auth_providers/github_app.py:76: error: Incompatible return value type (got "None", expected "str")
ogr/services/github/auth_providers/github_app.py:87: error: unused 'type: ignore' comment
[...]

PR #256, opened 2019-10-26
Improve typehints for exceptions.py

svenharris:
This pull request intends to fix the strict mypy errors in the exceptions.py file as part of #251. Not sure if the types within the dictionary (str, str) are consistent with how this is used as couldn't see any instances in the codebase where this error is raised with the kwargs filled in.

### DIFF
--- a/ogr/exceptions.py
+++ b/ogr/exceptions.py
@@ -30,7 +30,10 @@ class PagureAPIException(OgrException):
     """ Exception related to Pagure API """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


ogr/exceptions.py
```diff
@@ -29,7 +30,10 @@ class PagureAPIException(OgrException):
     """ Exception related to Pagure API """
 
     def __init__(
-        self, *args: object, pagure_error: str = None, pagure_response: dict = None
+        self,
+        *args: object,
```
lachmanfrantisek: What about `Any`?

lachmanfrantisek:
Thanks!

PR #258, opened 2019-10-26
Improve type hint coverage in abstract.py

svenharris:
As part of issue #251 update type hinting within the abstract.py file such that there are no mypy errors for that file

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -22,7 +22,7 @@
 
 import datetime
 from enum import IntEnum
-from typing import Optional, Match, List, Dict, Set, TypeVar, Any, NoReturn
+from typing import Optional, Match, List, Dict, Set, TypeVar, Any
 from urllib.request import urlopen
 
 from ogr.parsing import parse_git_repo
@@ -226,11 +226,11 @@ class Release:
         self.project = project
 
     @property
-    def title(self) -> NoReturn:
+    def title(self) -> str:
         raise NotImplementedError()
 
     @property
-    def body(self) -> NoReturn:
+    def body(self) -> str:
         raise NotImplementedError()
 
     def save_archive(self, filename: str) -> None:
@@ -252,7 +252,7 @@ class Release:
             f"tarball_url='{self.tarball_url}')"
         )
 
-    def edit_release(self, name: str, message: str) -> NoReturn:
+    def edit_release(self, name: str, message: str) -> None:
         """
         Edit name and message of a release.
 
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -22,7 +22,7 @@
 
 import datetime
 from enum import IntEnum
-from typing import Optional, Match, List, Dict, Set, TypeVar, Any
+from typing import Optional, Match, List, Dict, Set, TypeVar, Any, Sequence
 from urllib.request import urlopen
 
 from ogr.parsing import parse_git_repo
@@ -791,7 +791,7 @@ class GitProject:
         """
         raise NotImplementedError
 
-    def get_forks(self) -> List["GitProject"]:
+    def get_forks(self) -> Sequence["GitProject"]:
         """
         Get forks of the project.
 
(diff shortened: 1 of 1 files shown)


svenharris:
Something failing here still on the type hinting in the zuul. May be a MyPy version thing vs what I have locally, will rerun using the same precommit config as in the repo and check once I get a chance

ogr/abstract.py
```diff
@@ -281,6 +280,7 @@ def get_project(self, **kwargs) -> "GitProject":
 
     def get_project_from_url(self, url: str) -> "GitProject":
         repo_url = parse_git_repo(potential_url=url)
+        assert repo_url, f"Failed to return repo for url: {url}"
```
lachmanfrantisek: The assert should not be in the real code -- it is meant for testing and development. (The code can be run in such way, that the `assert` statements are skipped.)

ogr/abstract.py
```diff
@@ -495,7 +497,7 @@ def issue_close(self, issue_id: int) -> Issue:
         """
         raise NotImplementedError()
 
-    def get_issue_labels(self, issue_id: int) -> List:
+    def get_issue_labels(self, issue_id: int) -> List[Any]:
         """
         Get list of issue's labels.
```
lachmanfrantisek: This is what we need to fix -- we should have the same return value for all implementations.

Can you please fix that or create an issue describing the problem?

With "fix" I mean the same structure of the output for all implementations. (Probably using a new class for labels.)

ogr/abstract.py
```diff
@@ -732,15 +743,15 @@ def pr_merge(self, pr_id: int) -> "PullRequest":
         """
         raise NotImplementedError()
 
-    def get_pr_labels(self, pr_id: int) -> List:
+    def get_pr_labels(self, pr_id: int) -> List[Any]:
```
lachmanfrantisek: Same as above.

ogr/abstract.py
svenharris: Created issue #263, what do you want doing in the case of this pull request? Revert these lines or happy to leave as Any?

ogr/abstract.py
svenharris: Do you want me to remove this line? I put it in for type hinting reasons because output from `parse_git_repo` can be `None` (unless we assert that it isn't) and so mypy complains that we're trying to access the attributes of a NoneType in the line below this. Without the assertion I think you would get an AttributeError if the parsing returned a None, which seems to be a valid path. 

Or did you mean that a different error type should be thrown here in the None case so that the assertion cannot be skipped?

lachmanfrantisek: Thanks for creating the issue -- let's leave it as is (with `Any`).

ogr/abstract.py
lachmanfrantisek: Thanks for creating the issue -- let's leave it as is (with Any).

ogr/abstract.py
lachmanfrantisek: Ideally, we want to fix that -- test the `repo_url` to `not None` and raise an `OgrException` otherwise.

(So we can remove the `assert`.)

lachmanfrantisek:
recheck

lachmanfrantisek:
> Zuul merged the master, which means your branch is not up to date with upstream/master.
> Please, rebase!

ogr/abstract.py
```diff
@@ -281,6 +281,8 @@ def get_project(self, **kwargs) -> "GitProject":
 
     def get_project_from_url(self, url: str) -> "GitProject":
         repo_url = parse_git_repo(potential_url=url)
+        if not repo_url:
+            raise OgrException(f"Failed to find repository for url: {url}")
```
lachmanfrantisek: :+1:

lachmanfrantisek:
One more rebase will be needed since we merged your other PR (#259). Thanks!

lachmanfrantisek:
Thanks!

PR #259, opened 2019-10-26
improve typehint coverage in utils.py

svenharris:
Aims to remove any strict MyPy errors from the utils.py file, relating to ticket #251

### DIFF
--- a/ogr/utils.py
+++ b/ogr/utils.py
@@ -24,7 +24,7 @@ import logging
 import os
 import re
 import subprocess
-from typing import List, Union, Match, Optional
+from typing import List, Union, Match, Optional, Dict, Tuple, Any
 
 from ogr.abstract import AnyComment, Comment
 from ogr.constant import CLONE_TIMEOUT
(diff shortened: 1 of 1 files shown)


ogr/utils.py
```diff
@@ -164,7 +166,7 @@ def __str__(self) -> str:
             f"RequestResponse("
             f"status_code={self.status_code}, "
             f"ok={self.ok}, "
-            f"content={self.content}, "
+            f"content={self.content.decode()}, "
```
lachmanfrantisek: @jscotka is this ok? (I hope we don't break anything related to requre.)

jscotka: should be okay requre should be trasparent.

lachmanfrantisek: jscotka said it should be OK..

lachmanfrantisek:
recheck

lachmanfrantisek:
Please, rebase so we can merge that.

PR #260, opened 2019-10-27
Comment: mypy

mfocko:
- Update annotation of `_from_raw_comment`

### DIFF
--- a/ogr/services/pagure/comments.py
+++ b/ogr/services/pagure/comments.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 
 import datetime
-from typing import Optional
+from typing import Optional, Dict, Any
 
 from ogr.abstract import IssueComment, PRComment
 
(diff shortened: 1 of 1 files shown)


tests/integration/test_github.py
```diff
@@ -80,6 +80,12 @@ def test_pr_comments(self):
         assert pr_comments[0].comment.endswith("fixed")
         assert pr_comments[1].comment.startswith("LGTM")
 
+        comment_str = str(pr_comments[0])
+        assert comment_str.startswith(
+            "GithubPRComment(comment='@TomasTome...', author='lachmanfrantisek', created='"
+        ) and comment_str.endswith(")")
+        assert "edited" in comment_str
```
lachmanfrantisek: I am not sure if these tests are useful, but why not. (I understand the coverage effect, but practically it doesn't gives us anything practical.)

mfocko: Up to you, it's just regarding the coverage, I can drop it.

PR #262, opened 2019-10-31
Add config for stale bot

lachmanfrantisek:
The used config is the example one from the app page:
https://github.com/apps/stale/

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false


lachmanfrantisek:
More options can be found here:
https://github.com/probot/stale

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
```
TomasTomecek: not sure if I'm picking nits but I'd start with more benevolent configuration: 90 days of inactivity and 14 days to close.

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
TomasTomecek: how 'bout adding triaged here?

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
```
jpopelka: Not sure I understand the `closeComment`.
Does it mean that when I set it to `true` a comment will be posted when a stale issue is closed?
But what comment?

.github/stale.yml
lachmanfrantisek: OK, sounds reasonable.

.github/stale.yml
lachmanfrantisek: Currently, most of the issues are marked as triaged so the bot wil do nothing in that case.

.github/stale.yml
TomasTomecek: "This issue was exterminated!"

.github/stale.yml
jpopelka: We'd need to redefine how we use it because currently it means "someone has taken a look at it" - so we add it to basically any issue.

.github/stale.yml
lachmanfrantisek: I think it's as you write, but I don't understand that as well. There is no documentation with the explanation -- I only copied their example.

We can try the default one and tweak it if needed.

.github/stale.yml
lachmanfrantisek: So. What about removing `triaged` from all issues?

PR #275, opened 2019-11-14
temporary integration test method PullRequest._pr_close_temp() removed

sakalosj:


sakalosj:
is it possible to disable triggering of checks/builds on test branches test_source and test_target in hello-world repository?

lachmanfrantisek:
> is it possible to disable triggering of checks/builds on test branches test_source and test_target in hello-world repository?

We cannot blacklist the branches, but you can try to remove the `packit.yaml` config file so the build should not be triggered.

We are using that repo for testing the Packit-as-a-Service so we need to have the app(s) enabled.

sakalosj:
> > is it possible to disable triggering of checks/builds on test branches test_source and test_target in hello-world repository?
> 
> We cannot blacklist the branches, but you can try to remove the `packit.yaml` config file so the build should not be triggered.
> 
> We are using that repo for testing the Packit-as-a-Service so we need to have the app(s) enabled.

ok

sakalosj:
i have removed .packit.yaml and build is not triggered anymore.

@lachmanfrantisek if there is nothing else i thinnk it can be merged

lachmanfrantisek:
> @lachmanfrantisek if there is nothing else i thinnk it can be merged

Thanks!

PR #279, opened 2019-11-20
Add deprecation policy and dependency package

mfocko:


mfocko:
Not sure if I should add Deprecated to `recipe.yaml` too

tox.ini
```diff
@@ -9,6 +9,7 @@ deps =
     pytest
     pytest-cov
     git+https://github.com/packit-service/requre
+    Deprecated
```
lachmanfrantisek: Here, we need to specify test-only packages. => Package `Deprecated` should be installed with the package itself.
(Try removing the tox cache directory if you have some problems with that.)

mfocko:
OK, and what about the `recipe.yaml`?

lachmanfrantisek:
> OK, and what about the `recipe.yaml`?

Yes, you can add that here as well (`PyYAML` is missing as well).

But it's not used a lot. We are mostly using playbooks/tasks in the `./files` directory.

mfocko:
So add to this `zuul-install-requirements-pip.yaml` then?

TomasTomecek:
```
functions will be removed after 6 months from the release in which they were marked as deprecated
```
this is definitely not enough: the common practice is to deprecate functionality and remove it in n+3 or n+5 release, examples:

https://docs.djangoproject.com/en/dev/internals/release-process/#internal-release-deprecation-policy
https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#deprecation-cycle

please update the line in the research to state something: "mark the functionality as deprecated and clearly state in which release it's going to be removed"

PR #286, opened 2019-11-30
Commit flags

mfocko:
- [x] add methods
- [x] services' implementation
- [x] enum for statuses?
- [x] PRs?
  - [x] tests

Comparison of statuses:

GitHub | GitLab | Pagure
--- | --- | ---
`pending` | `pending` | `pending`
`success` | `success` | `success`
`failure` | `failed` | `failure`
`error` | `-` | `error`
`-` | `canceled` | `canceled`
`-` | `running` | `-`

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -530,6 +530,7 @@ class CommitFlag:
             self.url = url
         else:
             self._raw_commit_flag = raw_commit_flag
+            self._from_raw_commit_flag()
         self.uid = uid
         self.project = project
 
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 
 import datetime
-from enum import IntEnum
+from enum import Enum, IntEnum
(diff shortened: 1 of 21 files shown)


ogr/abstract.py
```diff
@@ -437,35 +437,84 @@ def add_label(self, *labels: str) -> None:
...
+    def get_statuses(self) -> List["CommitFlag"]:
+        """
+        Returns statuses for latest commit on pull request.
+
+        :return: [CommitFlag]
+        """
+        raise NotImplementedError()
+
+
+class CommitStatus(IntEnum):
```
lachmanfrantisek: Do we need the comparison with `int`s?

What about this?
https://docs.python.org/3/library/enum.html#using-a-descriptive-string

mfocko: Had a look around abstract, I thought it's a convention, ok, I'll change it

lachmanfrantisek: It allows some `string <-> enum` conversion.

PR #287, opened 2019-12-02
Fix backward compatibility on pull requests

mfocko:


ogr/abstract.py
```diff
@@ -231,45 +231,70 @@ class PRStatus(IntEnum):
 
 
 class PullRequest:
-    def __init__(self, raw_pr: Any, project: "GitProject") -> None:
-        self._raw_pr = raw_pr
-        self.project = project
+    from ogr.utils import deprecate_and_set_removal as _deprecate_and_set_removal
```
lachmanfrantisek: What about having the definition in `deprecation.py` to avoid this.

Issue #288, opened 2019-12-03
AttributeError:  'GithubPRComment' object has no attribute '_body'


jpopelka:
Any ideas how [this](https://sentry.io/organizations/red-hat-0p/issues/1369231585) could have happen?

```python
AttributeError: 'GithubPRComment' object has no attribute '_body'
  File "packit_service/worker/tasks.py", line 42, in process_message
    return SteveJobs().process_message(event=event, topic=topic)
  File "packit_service/worker/jobs.py", line 249, in process_message
    jobs_results[job_type] = handler.run_n_clean()
  File "packit_service/worker/handler.py", line 177, in run_n_clean
    return self.run()
  File "packit_service/worker/fedmsg_handlers.py", line 221, in run
    if not self.was_last_build_successful():
  File "packit_service/worker/fedmsg_handlers.py", line 189, in was_last_build_successful
    if "Congratulations!" in comment.comment:
  File "ogr/abstract.py", line 78, in comment
    return self._body
```

PR #289, opened 2019-12-03
Fix typo in comments

mfocko:
Fixes: #288


jpopelka:
Not sure it fixes #288 since you now return `self.body`, which is just a property returning `self._body`, which the traceback claims did not exist at the moment of the crash. But I won't object since Franta ACKed it :-)

mfocko:
> Not sure it fixes #288 since you now return `self.body`, which is just a property returning `self._body`, which the traceback claims did not exist at the moment of the crash. But I won't object since Franta ACKed it :-)

the `body` property is overriden in service-specific comments

PR #293, opened 2019-12-04
Change statuses from `open` to `opened`

### LINKED ISSUE #193: Investigate 'open' x 'opened' status for Issues and PullRequests
We are using `open` in the status enums, but `closed`.

Currently, we need to replace that for GitLab manually and GitHub probably needs that fix as well. I am not sure about Pagure.

Ideally, we need to change our API, but we have some usages in other projects:

- [ ] packit
    - https://github.com/packit-service/packit/blob/master/tests/testsuite_basic/conftest.py#L114 (just a test)
- [x] packit-service (no occurences found)
- [ ] release-bot
    - https://github.com/user-cont/release-bot/blob/master/release_bot/github.py#L308

mfocko:
Closes #193

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -265,7 +265,8 @@ class Issue:
 
 
 class PRStatus(IntEnum):
-    open = 1
+    opened = 1
+    open = opened
     closed = 2
     merged = 3
     all = 4
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -117,7 +117,8 @@ class PRComment(Comment):
 
 
 class IssueStatus(IntEnum):
-    open = 1
+    opened = 1
+    open = opened
     closed = 2
     all = 3
 
--- a/ogr/services/github/issue.py
+++ b/ogr/services/github/issue.py
@@ -22,6 +22,7 @@
 
 import datetime
 from typing import List
+import warnings
 
 from github import UnknownObjectException
 from github.Issue import Issue as _GithubIssue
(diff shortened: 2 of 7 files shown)


ogr/abstract.py
```diff
@@ -117,7 +117,8 @@ def __str__(self) -> str:
 
 
 class IssueStatus(IntEnum):
-    open = 1
+    opened = 1
+    open = opened
```
lachmanfrantisek: Another reason, why `open` is not well named is that it shadows the built-in `open` function.

ogr/services/github/issue.py
```diff
@@ -93,10 +98,20 @@ def get(project: "ogr_github.GithubProject", id: int) -> "Issue":
 
     @staticmethod
     def get_list(
-        project: "ogr_github.GithubProject", status: IssueStatus = IssueStatus.open
+        project: "ogr_github.GithubProject", status: IssueStatus = IssueStatus.opened
     ) -> List["Issue"]:
+        if status == IssueStatus.open:
+            warnings.warn(
+                "Using deprecated constant, that will be removed in 0.14.0"
+                "(or 1.0.0 if it comes sooner). Please use opened.",
```
lachmanfrantisek: - `constant`  -> `enum value`
- `opened` -> `IssueStatus.opened`

lachmanfrantisek:
After the discussion, we agreed on not doing that. This naming is not used anywhere. (I'll close the related issue as well.)

PR #297, opened 2019-12-17
Add method to list files

lbarcziova:


ogr/abstract.py
```diff
@@ -1230,6 +1230,14 @@ def get_file_content(self, path: str, ref: str = "master") -> str:
         """
         raise NotImplementedError
 
+    def get_files(self, ref: str = "master") -> List[str]:
+        """
+        Get files of the project.
+
+        :return: [str]
```
lachmanfrantisek: From the method name and dostring, I would not be sure, what will I get.

- Mentioning the `path` would be useful.
- Do I get all of the files or only the top directory,...?

ogr/services/github/project.py
```diff
@@ -373,6 +373,26 @@ def get_file_content(self, path: str, ref="master") -> str:
...
             raise FileNotFoundError(f"File '{path}' on {ref} not found", ex)
 
+    def get_files(self, ref: str = "master") -> List[str]:
+        """
+        Get files of the project.
+
+        :return: [str]
+        """
+        files = []
+        contents = self.github_repo.get_contents(path="", ref=ref)
```
lachmanfrantisek: Some filtering and/or level limit would be nice. What do you think?

ogr/services/gitlab/project.py
```diff
@@ -326,6 +326,19 @@ def get_file_content(self, path, ref="master") -> str:
...
             raise FileNotFoundError(f"File '{path}' on {ref} not found", ex)
 
+    def get_files(self, ref: str = "master") -> List[str]:
+        """
+        Get files of the project.
+
+        :return: [str]
+        """
+        return [
+            file["path"]
```
lachmanfrantisek: It looks like `file` is some special keyword.

ogr/services/github/project.py
```diff
@@ -373,6 +373,26 @@ def get_file_content(self, path: str, ref="master") -> str:
...
+
+        while contents:
+            file_content = contents.pop(0)
+            if file_content.type == "dir":
+                contents.extend(
+                    self.github_repo.get_contents(path=file_content.path, ref=ref)
```
lachmanfrantisek: I am not sure about the pygithub/python-gitlab internals, but I want to be sure that we are not downloading the file contents. (If it is doable.)

lbarcziova:
@lachmanfrantisek I haven't found any reference to listing files in Pagure API yet. Are we going to merge and possibly add the Pagure method later?

lachmanfrantisek:
> @lachmanfrantisek I haven't found any reference to listing files in Pagure API yet. Are we going to merge and possibly add the Pagure method later?

If we don't currently need that, create an issue for it and we can go on with github/gitlab only for now.

lbarcziova:
In addition, have you already looked into the requre problem? It still does not work for me to generate the response files.

lbarcziova:
Since the test responses are generated now, can we merge?

Issue #214, opened 2019-09-19
[question] Allow objects representing nonexisting objects


lachmanfrantisek:
Currently, we can instantiate objects (i.e. GitProject), that does not exist.

We probably need to agree on the common understanding of this approach:

1. allow creating non-existing objects
    - add `exists() -> Bool` method
    - add `create() -> ????` methods
    - we do not need to connect to servers on each instantiation
    - due to laziness it can fail at not-predictable places
2. raise an exception in methods like `get_project`
    - easier implementation
    - we need to connect to the remote service even when it is not needed

lachmanfrantisek:
It's also about documenting these "principles"...

rpitonak:
Add parameter `verify=True` to `get_project` method. We would be still able to create "fake projects".

lachmanfrantisek:
We need to at least clarify the outcome from this.

mfocko:
did we do something about this in the end?

lachmanfrantisek:
I don't think so. Do you think we need to take a look? As I wrote above, at least some agreement and, ideally, docs in the contribution guide would be nice.

PR #311, opened 2020-01-24
Add compatibility table

mfocko:
- [ ] should be thoroughly checked (`Project`, `Service`, `User`, `Release` should be fine)

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,0 +1,65 @@
+# Compatibility table for services
+
+Methods/properties not listed in here are implemented for all services.
+In case of an error, please report with issue.
+
+## Comments
+
+### `IssueComment`
+
+|               | GitHub | GitLab | Pagure |
+| ------------- | :----: | :----: | :----: |
+| `body`-setter |   ✔    |   ✔    |   ✘    |
+
+### `PRComment`
+
+|               | GitHub | GitLab | Pagure |
+| ------------- | :----: | :----: | :----: |
+| `body`-setter |   ✔    |   ✔    |   ✘    |
+
+## Issue
+
+|             | GitHub | GitLab | Pagure |
+| ----------- | :----: | :----: | :----: |
+| `add_label` |   ✔    |   ✔    |   ✘    |
+
+## Pull request
+
+|                   | GitHub | GitLab | Pagure |
+| ----------------- | :----: | :----: | :----: |
+| `add_label`       |   ✔    |   ✔    |   ✘    |
+| `get_all_commits` |   ✔    |   ✔    |   ✘    |
+| `labels`-getter   |   ✔    |   ✔    |   ✘    |
+| `patch`           |   ✘    |   ✘    |   ✔    |
+
+## Release
+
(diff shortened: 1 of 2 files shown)


TomasTomecek:
Would be mindblowing if this was generated :P

TomasTomecek:
> Only collaborators can trigger Packit-as-a-Service...

isn't it time to make @mfocko a collaborator?

mfocko:
> Would be mindblowing if this was generated :P

I've thought about this... There are few problems with that:
1. some classes have their own members, e.g. `GithubProject` has `github_repo` or `GitlabProject` has `gitlab_repo`, which makes a bit of a mess
2. we could probably create objects and try to call all methods/properties from interface (`abstract.py` in this case) and "expect" exception
    - seems related to #214 
    - constructors for service-specific classes expect objects from the library (or JSON response in case of Pagure), so their creation could be a problem
    - checking methods that modify state or setters could be a problem too
3. have some attribute on methods? (properties are problem already because of decorators)
    - also if it's to be automated as much as it can, there is a problem how to change it without (in the best case) any repetition of code
    - I've noticed one method in `GithubProject` that raises `NotImplementedError` => another problem

lachmanfrantisek:
stalebot, we need to discuss that a bit.

lachmanfrantisek:
@mfocko What do we do with this PR?

What about updating the tables and merging? The automation can be done later.

mfocko:
> What about updating the tables and merging? The automation can be done later.

Sure, I'll have a look at it

lachmanfrantisek:
> Sure, I'll have a look at it

Thanks!

COMPATIBILITY.md
```diff
@@ -0,0 +1,65 @@
+# Compatibility table for services
+
+Methods/properties not listed in here are implemented for all services.
+In case of an error, please report with issue.
```
lachmanfrantisek, line 4:
```suggestion
In case of an error, please [create a new issue](https://github.com/packit-service/ogr/issues/new).
```

COMPATIBILITY.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Comments
+
+### `IssueComment`
+
+|               | GitHub | GitLab | Pagure |
+| ------------- | :----: | :----: | :----: |
+| `body`-setter |   ✔    |   ✔    |   ✘    |
```
lachmanfrantisek, line 12: It is also getter, isn't it? It's not clear what the cross means. Does it mean that Pagure implementation does not have `body` at all?

What about two lines or `get`/`set` in the parentheses?

README.md
```diff
@@ -53,6 +53,11 @@ This will output:
 
 You can use the same API for other forges, you just need to replace `GithubService` with `PagureService`.
 
+## Supported functionality
+
+For more info on functionality that _is not_ supported in all services the same way
+see [compatibility table](/COMPATIBILITY.md).
```
lachmanfrantisek, line 59:
```suggestion
see [compatibility tables](/COMPATIBILITY.md).
```

PR #317, opened 2020-01-28
github.pr.author: use login instead of name

TomasTomecek:
TomasTomecek vs Tomas Tomecek


TomasTomecek:
what, no CI failure?! let me add one assert for this

TomasTomecek:
recheck

PR #319, opened 2020-01-31
Remove trailing slash from URLs before parsing

nickcannariato:
Hello there 👋 

This came up in my `good-first-issue` search and seemed like a fairly straightforward bug to fix.

I decided to treat it as a data problem and check for the existence of a trailing slash in a simple if statement. It worked in the test cases I passed it, and should be pretty straightforward to debug. 

Open to any/all criticism. Thanks for all you do!

### DIFF
--- a/ogr/parsing.py
+++ b/ogr/parsing.py
@@ -90,6 +90,10 @@ def parse_git_repo(potential_url: str) -> Optional[RepoUrl]:
     if not potential_url:
         return None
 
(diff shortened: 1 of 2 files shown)


nickcannariato:
☝️ force pushed 4cd9543 & 7344695 because `black` was apparently not working in my editor when I made the commit. AFAIK it was only whitespace changes.

Should pass the tests now.

PR #320, opened 2020-01-31
Add `.gitignore` to repo

nickcannariato:
While working on #319, I noticed that there was no `.gitignore` file in the repository.

To solve my _own_ pain points while developing, I generated this one, which should ignore standard files for MacOS, Windows, & Linux, and additionally ignore files that are generally good to ignore in Python projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,273 @@
+# Created by https://www.gitignore.io/api/code,python,pycharm,macos,windows,linux
+# Edit at https://www.gitignore.io/?templates=code,python,pycharm,macos,windows,linux
+
+### Code ###
+.vscode
+!.vscode/settings.json
+!.vscode/tasks.json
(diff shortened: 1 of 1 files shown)


nickcannariato:
> please rebase to pass the pre-commit check

@TomasTomecek Done! Should pass the pre-commit check now and be good to go. 🙇

nickcannariato:
> Sorry for the nit-picking

No need to apologize @lachmanfrantisek! This definitely fell off my radar with work last week so I appreciate the ping.

Just went through and cleaned things up. Everything passes now and the amended commit should be good to merge. Let me know if I missed a rebase or something, otherwise this should be set and will ignore common, unnecessary files generated by:

- VS Code
- Python
- PyCharm
- Macos
- Windows
- *nix

Let me know if there's anything else that needs to be done here!

PR #321, opened 2020-01-31
Add support for tags when creating pagure issues.

cverna:
This commit allows to pass a list of tags to pagure's create_issue method.

Signed-off-by: Clement Verna <cverna@tutanota.com>

### DIFF
--- a/ogr/services/pagure/project.py
+++ b/ogr/services/pagure/project.py
@@ -208,8 +208,10 @@ class PagureProject(BaseGitProject):
     def get_issue(self, issue_id: int) -> Issue:
         return PagureIssue.get(project=self, id=issue_id)
 
-    def create_issue(self, title: str, body: str) -> Issue:
-        return PagureIssue.create(project=self, title=title, body=body)
+    def create_issue(
+        self, title: str, body: str, labels: Optional[List[str]] = None
+    ) -> Issue:
+        return PagureIssue.create(project=self, title=title, body=body, labels=labels)
 
     def get_pr_list(
         self, status: PRStatus = PRStatus.open, assignee=None, author=None
--- a/ogr/services/pagure/issue.py
+++ b/ogr/services/pagure/issue.py
@@ -74,12 +74,23 @@ class PagureIssue(BaseIssue):
     def created(self) -> datetime.datetime:
         return datetime.datetime.fromtimestamp(int(self._raw_issue["date_created"]))
 
+    @property
+    def labels(self) -> List[str]:
+        return self._raw_issue["tags"]
+
     def __str__(self) -> str:
         return "Pagure" + super().__str__()
 
     @staticmethod
-    def create(project: "ogr_pagure.PagureProject", title: str, body: str) -> "Issue":
+    def create(
+        project: "ogr_pagure.PagureProject",
+        title: str,
+        body: str,
+        labels: Optional[List[str]] = None,
+    ) -> "Issue":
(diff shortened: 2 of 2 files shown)


cverna:
I believe that would solve https://github.com/packit-service/ogr/issues/147

ogr/services/pagure/project.py
```diff
@@ -208,8 +208,10 @@ def get_issue_list(
     def get_issue(self, issue_id: int) -> Issue:
         return PagureIssue.get(project=self, id=issue_id)
 
-    def create_issue(self, title: str, body: str) -> Issue:
-        return PagureIssue.create(project=self, title=title, body=body)
+    def create_issue(
+        self, title: str, body: str, tags: Optional[List[str]] = None
```
cverna: @lachmanfrantisek would you also change `tags` here for `labels` so that it is consistent ?

lachmanfrantisek: Yes, that would be nice.

lachmanfrantisek:
I have three possible extensions of this PR: (No pressure, we can fix that later..;)

1. In other implementations, we have also `add_label` method, do you want to add that as well?

- gitlab: https://github.com/packit-service/ogr/blob/master/ogr/services/gitlab/issue.py#L136-L139
- github: https://github.com/packit-service/ogr/blob/master/ogr/services/github/issue.py#L139-L141
- and ideally add it to `abstract.py` since it is not there

2. Do you want to add the same code for the `PullRequest` class?
3. Do you want to help with adding some tests? (It's not so easy to start with our tests but after some experience it's very easy to create and use such tests...) -- See [our CONTRIBUTION guide](https://github.com/packit-service/ogr/blob/master/CONTRIBUTING.md#testing) and the other tests for more info.

Thanks!

TomasTomecek:
please rebase :)

cverna:
> I have three possible extensions of this PR: (No pressure, we can fix that later..;)
> 
>     1. In other implementations, we have also `add_label` method, do you want to add that as well?

So I think this is currently not supported in pagure. The API supports adding labels when creating an issue but changing labels while editing an issue does not work.

> 
> 
>     * gitlab: https://github.com/packit-service/ogr/blob/master/ogr/services/gitlab/issue.py#L136-L139
> 
>     * github: https://github.com/packit-service/ogr/blob/master/ogr/services/github/issue.py#L139-L141
> 
>     * and ideally add it to `abstract.py` since it is not there
> 
> 
>     1. Do you want to add the same code for the `PullRequest` class?

I can have a look, in another PR maybe

> 
>     2. Do you want to help with adding some tests? (It's not so easy to start with our tests but after some experience it's very easy to create and use such tests...) -- See [our CONTRIBUTION guide](https://github.com/packit-service/ogr/blob/master/CONTRIBUTING.md#testing) and the other tests for more info.

I had a quick look at it, but did not invest much time there tbh.

just to let you know I ll be on PTO for a week, so I won't be much around the keyboard.

> 
> 
> Thanks!

lachmanfrantisek:
@cverna thanks for the info!

> So I think this is currently not supported in pagure. The API supports adding labels when creating an issue but changing labels while editing an issue does not work.

I'll update the original issue to not forget about it.

> I can have a look, in another PR maybe

Nice, I'll add a note about that to the issue as well so we can merge this one now.

> I had a quick look at it, but did not invest much time there tbh.

Let's leave it as is for now.

PR #328, opened 2020-02-17
provide is_private method on GitProjects

dhodovsk:


ogr/services/pagure/project.py
```diff
@@ -291,6 +291,19 @@ def exists(self):
...
+        Is this repo private? (accessible only by users with granted access)
+
+        :return: if yes, return True
+        """
+        if self.service.instance_url == "https://src.fedoraproject.org":
+            # private repositories are not allowed in fedora
+            return False
+        raise NotImplementedError(
+            f"is_private is not implemented for {self.service.instance_url}"
+        )
```
lachmanfrantisek: - A little hacky..;) => Can you please create an issue for the real implementation? (So we can discuss if it is possible to implement it.)
- I am not sure about the exact match of `"https://src.fedoraproject.org"`.
  - Will the instance URL always be with `https`? If you check that from packit, I am ok with merging as is. I am just not 100% sure...

tests/integration/test_github.py
```diff
@@ -359,6 +359,9 @@ def test_get_web_url(self):
     def test_full_repo_name(self):
         assert self.ogr_project.full_repo_name == "packit-service/ogr"
 
+    def test_is_private(self):
+        assert not self.ogr_project.is_private()
```
lachmanfrantisek: Is it possible to test some private repository? Is it possible to create one in our organisation?

tests/integration/test_gitlab.py
```diff
@@ -172,6 +172,9 @@ def test_pr_permissions(self):
         assert self.project.can_merge_pr("lachmanfrantisek")
         assert not self.project.can_merge_pr("not_existing_user")
 
+    def test_is_private(self):
+        assert not self.project.is_private()
```
lachmanfrantisek: Same as above.

tests/integration/test_github.py
dhodovsk: Our current plan does not support private repositories. We can create it on one of our users, but while re-generating the test data, the person would have to provide their own repository with their github key.

ogr/services/pagure/project.py
dhodovsk: `https://src.fedoraproject.org` is the default and also used in packit service.
I agree that it's hacky, but I was not able to find good info in pagure doc for obtaining this info. But if it's important to you, I'll contact some human.

tests/integration/test_github.py
lachmanfrantisek: OK, but we don't need to share the keys. Just create the repository with the same name and use that repository in tests. (The namespace should be the user of the token.)

tests/integration/test_gitlab.py
lachmanfrantisek: But in GitLab, we can create private repositories easily.

ogr/services/pagure/project.py
dhodovsk: I tried to make it a bit nicer and talked with pingou - here is the issue: https://github.com/packit-service/ogr/issues/330

lachmanfrantisek: Thanks!

ogr/services/pagure/project.py
```diff
@@ -291,6 +292,20 @@ def exists(self):
         response = self._call_project_api_raw()
         return response.ok
 
+    def is_private(self) -> bool:
+        """
+        Is this repo private? (accessible only by users with granted access)
+
+        :return: if yes, return True
+        """
+        host = urlparse(self.service.instance_url).hostname
```
lachmanfrantisek: We already have the parsing ability in `./parsing.py` that can be used instead, but  it's not so important.

```python
In [1]: from ogr.parsing import parse_git_repo                                        

In [2]: str(parse_git_repo("https://pagure.io"))                                      
Out[2]: "RepoUrl(repo='', namespace='None', is_fork=False, hostname='pagure.io', scheme='https')"
```

tests/integration/test_github.py
```diff
@@ -359,6 +359,16 @@ def test_get_web_url(self):
     def test_full_repo_name(self):
         assert self.ogr_project.full_repo_name == "packit-service/ogr"
 
+    def test_is_not_private(self):
+        # when regenerating this test with your gitlab token, use your own private repository
+        private_project = self.service.get_project(
+            namespace="dhodovsk", repo="playground"
+        )
+        assert private_project.is_private()
```
lachmanfrantisek: Thanks!

lachmanfrantisek:
>     * [ogr-pre-commit ](https://softwarefactory-project.io/zuul/t/local/build/76c744e8686e49b28b6cc569d969157b) : FAILURE in 3m 33s

Needs rebase...

dhodovsk:
yup, I can read :)

lachmanfrantisek:
> yup, I can read :)

Sorry.

Issue #331, opened 2020-02-18
pagure provides more detailed error info stored under 'errors' key


sakalosj:
pagure provides more detailed error info stored under 'errors' key, which is not currently not used in output

PR #333, opened 2020-02-18
improve pagure error response

sakalosj:
improve pagure error response to show additional info stored under 'errors' key


sakalosj:
fixes #331

sakalosj:
yes i can, will open separate issue for that

PR #336, opened 2020-02-20
Do not use tokens and keys in the __str__ methods

lachmanfrantisek:
- Do not use tokens/private_keys in the `__str__` methods.
  - used `set`/`not set` instead
  - reason: We've started to show logs publicly in the packit-service and it can be easy to forget about this "feature".

### DIFF
--- a/ogr/services/github/service.py
+++ b/ogr/services/github/service.py
@@ -30,8 +30,8 @@ from ogr.abstract import GitUser
 from ogr.exceptions import GithubAPIException
 from ogr.factory import use_for_service
 from ogr.services.base import BaseGitService
-from ogr.services.github.user import GithubUser
 from ogr.services.github.project import GithubProject
+from ogr.services.github.user import GithubUser
 
 
 @use_for_service("github.com")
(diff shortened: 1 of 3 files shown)


lachmanfrantisek:
Less secure, but more UX-friendly approach:

What about using only first letter(s) to be able to compare the tokens easily if I know them. (usecase: I want to check if there is a new or old token used.)

```python
PagureService(instance_url="https://pagure.io", token="abc...")
```

jpopelka:
> using only first letter(s)

agree, or maybe first and last letter with `***` between, e.g. `x***f`

lachmanfrantisek:
- updated to use first and the last letter
- changed the output to show flags only for non-default values

PTAL

ogr/services/github/service.py
```diff
@@ -77,12 +77,17 @@ def github_app_private_key(self):
         return None
 
     def __str__(self) -> str:
-        token_str = f", token='{self.token}'" if self.token else ""
+        token_str = (
+            f", token='{self.token[:1]}***{self.token[-1:]}'" if self.token else ""
```
jpopelka: those `:1` and `-1:` are clever, but do we need them (instead of more simple `0` and `-1`) when we have the `if self.token` ?

lachmanfrantisek: You are right. Do you want me to "fix! that in another PR?

PR #337, opened 2020-02-20
enable getting projects defined with SSH URLs

### LINKED ISSUE #306: ogr can't figure out auth for SSH style of URL: git@github...
see below: https://github.com/packit-service/ogr/issues/306#issuecomment-575204355

TomasTomecek:
Fixes #306

Franto, we didn't speak about this, but this how I expect it to work, WDYT?

### DIFF
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -81,10 +81,15 @@ def get_project(
     :return: GitProject using the matching implementation
     """
     kls = get_service_class(url=url, service_mapping_update=service_mapping_update)
+    repo_url = parse_git_repo(url)
 
     if custom_instances:
         for service_inst in custom_instances:
-            if isinstance(service_inst, kls) and service_inst.instance_url in url:
+            si_repo_url = parse_git_repo(service_inst.instance_url)
+            if (
+                isinstance(service_inst, kls)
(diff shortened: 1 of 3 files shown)


TomasTomecek:
will fix the tests when I get back

ogr/factory.py
```diff
@@ -81,10 +81,15 @@ def get_project(
     :return: GitProject using the matching implementation
     """
     kls = get_service_class(url=url, service_mapping_update=service_mapping_update)
+    repo_url = parse_git_repo(url)
 
     if custom_instances:
         for service_inst in custom_instances:
-            if isinstance(service_inst, kls) and service_inst.instance_url in url:
+            si_repo_url = parse_git_repo(repo_url.get_instance_url())
```
lachmanfrantisek: This is the second time we need to do it -- maybe a time to have a property for that in the service class.

(I'll create a good-first-issue.)

TomasTomecek:
I was actually really frustrated from the tests and wanted to delete them b/c of all the mocking, but then realized there is a real bug in my code :D so, well done, tests

lachmanfrantisek:
... rebase needed ... (sorry)

PR #345, opened 2020-03-04
Add stage to expected pagure instances

dhodovsk:


ogr/services/pagure/project.py
```diff
@@ -300,7 +300,7 @@ def is_private(self) -> bool:
         :return: if yes, return True
         """
         host = urlparse(self.service.instance_url).hostname
-        if host in ["src.fedoraproject.org", "pagure.io"]:
+        if host in ["src.fedoraproject.org", "pagure.io", "src.stg.fedoraproject.org"]:
             # private repositories are not allowed in src.fedoraproject.org or pagure.io
```
lachmanfrantisek: Can you please update the comment bellow. (Or change it so we don't need to duplicate the list...)

PR #351, opened 2020-03-11
Fix and refactor packit rev-dep tests

lbarcziova:
Related #350


files/tasks/packit-tests.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: install and run basic packit tests via ansible playbooks
+  command: ansible-playbook -e "ansible_python_interpreter=/usr/bin/python3" -v -c local -i localhost, files/zuul-tests.yaml
+  args:
+    chdir: "{{ reverse_dir }}"
+- name: install and run packit tests with session recording via ansible playbooks
+  command: ansible-playbook -e "ansible_python_interpreter=/usr/bin/python3" -v -c local -i localhost, files/zuul-tests-session-recording.yaml
+  args:
+    chdir: "{{ reverse_dir }}"
```
TomasTomecek: these are the playbooks coming from packit repo, right? I wonder how we could get rid of the nested `ansible-playbook` execution

lbarcziova: `these are the playbooks coming from packit repo, right? I` Yes.
` I wonder how we could get rid of the nested ansible-playbook execution` I will try to check more docs, maybe I will find something which can help us here

lbarcziova:
> Thanks!
> Just please test that we are installing the right ogr as you did in #350.

[Tested](https://github.com/packit-service/ogr/pull/351#issuecomment-597526924)

lachmanfrantisek:
> Tested

Thanks!

files/tasks/packit-tests.yaml
TomasTomecek: we can do import_tasks do run those tasks dynamically

lbarcziova: but the files which are called via ansible-playbook command are playbooks, not tasks

lbarcziova: I have tried  `import_playbook` and `include` options and both seem not to proceed `reverse_dir` variable :(

lbarcziova: so, can we merge this with ansible-playbook execution for now (as we were talking, it will be probably changed in the future)?

PR #352, opened 2020-03-16
github: raise when we didn't obtain install id

TomasTomecek:
~~Blocked on #357~~

Replaces an error like this:
```
github.GithubException.GithubException: 401 {"message":"'Issuer' claim ('iss') must be an Integer","documentation_url":"https://developer.github.com/v3"}
```

### DIFF
--- a/tests/integration/test_github_app.py
+++ b/tests/integration/test_github_app.py
@@ -30,7 +30,7 @@ class GithubTests(RequreTestCase):
             not self.github_app_id or not self.github_app_private_key_path
         ):
             raise EnvironmentError(
-                "You are in Requre write mode, please set"
+                "You are in Requre write mode, please set "
                 "GITHUB_APP_ID GITHUB_APP_PRIVATE_KEY_PATH env variables"
             )
 
--- a/tests/integration/test_github_app.py
+++ b/tests/integration/test_github_app.py
@@ -2,6 +2,11 @@ import os
 import tempfile
 from pathlib import Path
 
+import pytest
+from ogr.exceptions import OgrException
+
+from ogr.services.github.project import GithubProject
+
 from ogr import GithubService
 from requre.storage import PersistentObjectStorage
 from requre.utils import StorageMode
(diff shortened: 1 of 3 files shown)


tests/integration/test_github_app.py
```diff
@@ -79,3 +84,14 @@ def test_get_project_having_key_as_path(self):
         finally:
             if not self.github_app_private_key_path:
                 Path(github_app_private_key_path).unlink()
+
+    def test_github_proj_no_app_creds(self):
+        service = GithubService(
+            github_app_id="123", github_app_private_key=TESTING_PRIVATE_KEY
+        )
+        project = GithubProject("packit", service, "packit-service")
```
lachmanfrantisek: I wouůd use keyword arguments to be sure what's what...

TomasTomecek: You would, but I won't - this is perfectly clear to me:

https://github.com/packit-service/ogr/blob/5bc1dcf352e37c621d9a6a58225325fec85d0c21/ogr/services/github/project.py#L64

there is also the assert which makes sure the namespace and repo name are correct.

TomasTomecek:
I don't understand the test failures - they are passing for me locally and I have HEAD requre.

tests/integration/test_github_app.py
lachmanfrantisek: You need to check `__init__` to get that info. And the order of attributes is misleading.

I know what's what, but this is much clearer for anyone:

```suggestion
        project = GithubProject(service=service, namespace="packit-service", repo="packit")
```

lachmanfrantisek:
> I don't understand the test failures - they are passing for me locally and I have HEAD requre.

Different version of pygithub?

TomasTomecek:
> > I don't understand the test failures - they are passing for me locally and I have HEAD requre.
> 
> Different version of pygithub?

good point, let me retry w/ tox then

TomasTomecek:
nope

waiting for #353 to be fixed

tests/integration/test_github_app.py
csomh: The ordering of these arguments is a little bit weird: "service, namespace, repo" or "namespace, repo, service" would be much easier to guess. But given the current status +1 for being explicit with this in order to improve readability.

Issue #338, opened 2020-02-21
hostname property in the service classes


lachmanfrantisek:
Multiple times in the code (mostly in the `parsing.py`), we need to parse the hostname of the instance.

```python
service_repo_url = parse_git_repo(repo_url.get_instance_url())
service_repo_url.hostname
```

It would be useful to have a `hostame` property in service classes`.

TODO:
- [ ] add the (read-only) property to the `abstract.py`
- [ ] implement the common behaviour in the `base.py` (see the example above)
  - services with multiple instances possible
- [ ] hardcode the value for static instances (GitHub)
- [ ] use that property in places, where we are getting hostname of the instance
- [ ] create some unit tests for that

dinolinjob:
I would like to work on this issue

lachmanfrantisek:
@dinolinjob 
Nice, if you need some help, feel free to aks. The testing part is a bit tricky for the first time. (See our [contribution guide](https://github.com/packit-service/ogr/blob/master/CONTRIBUTING.md) for some info.)

PR #362, opened 2020-03-24
adding hostname property

dinolinjob:
During pre-commit I'm getting the following issues (Trying to solve #338 )
```
ogr/abstract.py:731: error: Too many arguments
ogr/services/github/service.py:172: error: Too many arguments
```


ogr/abstract.py
```diff
@@ -726,6 +727,13 @@ def project_create(
         """
         raise NotImplementedError()
 
+    @property
+    def get_hostname(self, url: str) -> str:
+        repo_url = parse_git_repo(potential_url=url)
+        service_repo_url = parse_git_repo(repo_url.get_instance_url())
+        hostname = service_repo_url.hostname
+        return hostname
```
lachmanfrantisek: 1. The properties cannot have an argument since they are used without parenthesis (e.g. `my_service.hostname`).
```suggestion
    @property
    def get_hostname(self) -> str:
        repo_url = parse_git_repo(potential_url=url)
        service_repo_url = parse_git_repo(repo_url.get_instance_url())
        hostname = service_repo_url.hostname
        return hostname
```

2. Use `self.instance_url` instead:

```suggestion
    @property
    def get_hostname(self) -> str:
        repo_url = parse_git_repo(potential_url=self.instance_url)
        service_repo_url = parse_git_repo(repo_url.get_instance_url())
        hostname = service_repo_url.hostname
        return hostname
```

3. The properties are used as attributes => the `get` does not make sense here:

```suggestion
    @property
    def hostname(self) -> str:
        repo_url = parse_git_repo(potential_url=self.instance_url)
        service_repo_url = parse_git_repo(repo_url.get_instance_url())
        hostname = service_repo_url.hostname
        return hostname
```


3.  We don't need to parse the url twice:

```suggestion
    @property
    def hostname(self) -> str:
        service_repo_url = parse_git_repo(potential_url=self.instance_url)
        hostname = service_repo_url.hostname
        return hostname
```

modules.out
```diff
@@ -0,0 +1,24 @@
+github.MainClass (/home/dinolin/Documents/Github/ogr/.tox/py36/lib/python3.6/site-packages/github/MainClass.py)-> requests (None)
```
TomasTomecek: and we don't want this one in git

dinolinjob: okay

ogr/abstract.py
```diff
@@ -726,6 +726,12 @@ def project_create(
         """
         raise NotImplementedError()
 
+    @property
+    def hostname(self) -> str:
+        service_repo_url = parse_git_repo(self.instance_url)
```
TomasTomecek: I'd say it would make sense to make the result of this function call to be a property of this class so it doesn't need to be called every time someone calls hostrname, what do you think?

ogr/services/github/service.py
```diff
@@ -166,3 +166,10 @@ def project_create(self, repo: str, namespace: str = None) -> "GithubProject":
             service=self,
             github_repo=new_repo,
         )
+
+    @property
+    def hostname(self) -> str:
+        service_repo_url = parse_git_repo(self.instance_url)
+        hostname = service_repo_url.hostname
+        return hostname
+
```
TomasTomecek: This code is identical to the one in the abstract class and will be inherited automatically, feel free to drop it.

TomasTomecek:
Glad you were able to resolve the issue with tests. The change looks good, I'd just cache the result of the parse function call.

We also need at least a single test case for this. Unit test is just fine.

Makefile
```diff
@@ -12,7 +12,7 @@ prepare-check:
 check:
 	@#`python3 -m pytest` doesn't work here b/c the way requre overrides import system:
 	@#`AttributeError: module 'importlib_metadata' has no attribute 'distributions'
-	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 pytest-3 --verbose --showlocals $(TEST_TARGET)
+	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 pytest --verbose --showlocals $(TEST_TARGET)
```
lachmanfrantisek: Please, do not include this change here. Create a new pull-request / issue for that so others can say something to that topic. (We want to solve one thing in one PR.)

dinolinjob: okay

ogr/abstract.py
dinolinjob: yeah..thanks for the suggestion

PR #367, opened 2020-03-26
add last_commit property to Pagure project

sakalosj:


ogr/services/pagure/pull_request.py
```diff
@@ -104,6 +104,10 @@ def created(self) -> datetime.datetime:
     def diff_url(self) -> str:
         return f"{self.url}#request_diff"
 
+    @property
+    def last_commit(self):
```
lachmanfrantisek: Maybe a `head_commit` would fit better. (Last is a bit misleading because I don't know the direction...)

sakalosj: changed

sakalosj:
> Please:
> 
> * [x]  Add this to the abstract service class with `NotImplementedError`.
> * [x]  Add a test(s) for this.
> * [x]  And ideally, implement this for all three implementations. (I hope it will be very easy.) Otherwise, create an issue for that.

yes should be easy, but still takes time, so i opened issue for that, possibly will look at it later

sakalosj:
> Thanks for the update.
> 
> To be honest, I would prefer using the same approach for tests as we do for everything else.
> 
> Just see the other existing tests how it's simple. Now, you actually reinventing the requre..;) I can help if you have some problems with that.

Its not reinventing requre, this is unit test it isolates tested method from all other stuff. In this case  method is using whole api reply stored in Pagure._raw_pr. It is enough to have reduced dictionary in test, but using full example you can just update `data_pagure_raw` fixture - replace value with parameter and use it other tests.

Requre is not unit testing tool by design.

lachmanfrantisek:
@sakalosj requre will do exactly this for you -- save the Pagure response. It will be messy to have multiple ways how to write a test. (Mainly, it will be misleading for newcomers.)



You can probably only add one line to this test since the response is already saved:

https://github.com/packit-service/ogr/blob/fe64b280c0fd5ef15d79991c233c67c9888ae486/tests/integration/test_pagure.py#L287-L297


So: one new line vs. 209 new lines and new requirement to test the same thing

sakalosj:
> @sakalosj requre will do exactly this for you -- save the Pagure response. It will be messy to have multiple ways how to write a test. (Mainly, it will be misleading for newcomers.)
> 
> You can probably only add one line to this test since the response is already saved:
> 
> https://github.com/packit-service/ogr/blob/fe64b280c0fd5ef15d79991c233c67c9888ae486/tests/integration/test_pagure.py#L287-L297
> 
> So: one new line vs. 209 new lines and new requirement to test the same thing

as mentioned before, Requre is not unit testing tool.

sakalosj:
added also Requere test

tests/integration/test_pagure.py
```diff
@@ -295,6 +295,7 @@ def test_pr_info(self):
             pr_info.diff_url
             == "https://pagure.io/ogr-tests/pull-request/5#request_diff"
         )
+        assert isinstance(pr_info.head_commit, str)
```
lachmanfrantisek: It does not differentiate the first/last commit.  (Or any other commit.) I would add the specific value as we do for all tests.

sakalosj:
> Thanks!

welcome, and will also remove unnecessary imports

PR #369, opened 2020-03-27
Added unit test for Pagure.PullRequest.head_commit

sakalosj:
Added unit test for Pagure.PullRequest.head_commit

### DIFF
--- a/files/tasks/rpm-test-deps.yaml
+++ b/files/tasks/rpm-test-deps.yaml
@@ -16,6 +16,7 @@
 - name: Install requre from git master
   pip:
     name:
+      - flexmock
       - git+https://github.com/packit-service/requre.git
     state: latest
   become: true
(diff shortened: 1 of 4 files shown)


sakalosj:
based on discussion with @lachmanfrantisek not required

TomasTomecek:
is this what we want to talk about in the arch sessions?

lachmanfrantisek:
> is this what we want to talk about in the arch sessions?

Originally yes, but we removed that. We have some idea about this approach with @sakalosj.

- [ ] have basic flexmock structure for all instances
- [ ] have some example tests
- [ ] have it documented in the contribution guide

Then, for each new functionality, you will:
- add a test using requre
- add a unit test using approach from this PR:
  - add a dummy value to the flexmock structure
  - add a test using the flexmock structure

Issue #344, opened 2020-03-02
Add datetime to commit flags


lachmanfrantisek:
- [x] add the `created` and `edited` properties to the commit flag classes
  - [x] abstract class
  - [x] github:
    - https://pygithub.readthedocs.io/en/latest/github_objects/Commit.html#github.Commit.Commit.get_statuses
  - [ ] gitlab:
    - https://python-gitlab.readthedocs.io/en/stable/gl_objects/commits.html#commit-status
    - https://docs.gitlab.com/ce/api/commits.html
  - pagure:
    -  https://src.fedoraproject.org/api/0/ -> `Flags for a commit`
- [x] tests for all implementations

TomasJani:
I would like to work on this if possible (as part of RHOSC).

lachmanfrantisek:
@TomasJani nice!

Feel free to ask if you need some help.

mfocko:
Closed by #374 

@lachmanfrantisek can create follow-up issue for the `edited` on GitLab, but I don't think it's the API problem

lachmanfrantisek:
Thanks. Yes please, create a follow-up for the missing parts. (Also issues blocked by API are useful and we shout create an issue in the upstream project for each of them.)

PR #374, opened 2020-04-08
Add datetime to commitflags

TomasJani:
Add datetime to commit flags #344

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -581,7 +581,9 @@ class CommitFlag:
             f"context='{self.context}', "
             f"uid='{self.uid}',"
             f"comment='{self.comment}',"
-            f"url='{self.url}')"
+            f"url='{self.url}'"
+            f"created='{self.created}'"
+            f"edited='{self.edited}')"
         )
 
     def _state_from_str(self, state: str) -> CommitStatus:
@@ -605,6 +607,14 @@ class CommitFlag:
         description: str,
         context: str,
     ) -> "CommitFlag":
+        raise NotImplementedError()
+
+    @property
+    def created(self) -> datetime.datetime:
+        raise NotImplementedError()
+
+    @property
+    def edited(self) -> datetime.datetime:
         raise NotImplementedError()
 
 
--- a/ogr/services/github/flag.py
+++ b/ogr/services/github/flag.py
@@ -19,6 +19,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import datetime
 import warnings
 from typing import List, Union
 
@@ -85,3 +86,11 @@ class GithubCommitFlag(CommitFlag):
             state.name, target_url, description, context
         )
         return GithubCommitFlag(project=project, raw_commit_flag=status, commit=commit)
+
+    @property
+    def created(self) -> datetime.datetime:
+        return self._raw_commit_flag.created_at
+
+    @property
+    def edited(self) -> datetime.datetime:
+        return self._raw_commit_flag.updated_at
--- a/ogr/services/gitlab/flag.py
+++ b/ogr/services/gitlab/flag.py
@@ -22,6 +22,7 @@
 
 import logging
 import warnings
+import datetime
 from typing import List, Union
 
 import gitlab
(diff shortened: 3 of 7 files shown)


tests/integration/test_pagure.py
```diff
@@ -501,6 +503,8 @@ def test_pr_status(self):
         assert statuses
         assert len(statuses) >= 0
         assert statuses[-1].state == CommitStatus.success
+        assert statuses[-1].created == datetime(year=2019, month=12, day=2, hour=14, minute=16, second=11)
```
lachmanfrantisek, line 506: Probably a typo:

```suggestion
        assert statuses[-1].created == datetime(year=2019, month=12, day=2, hour=13, minute=16, second=11)
```

See [the output](https://softwarefactory-project.io/zuul/t/local/build/f2589ccd20df4cec98bf27d3aaac2fe0/console) from zuul, our CI system:
```
>       assert statuses[-1].created == datetime(year=2019, month=12, day=2, hour=14, minute=16, second=11)
E       AssertionError: assert datetime.datetime(2019, 12, 2, 13, 16, 11) == datetime.datetime(2019, 12, 2, 14, 16, 11)
E        +  where datetime.datetime(2019, 12, 2, 13, 16, 11) = <ogr.services.pagure.flag.PagureCommitFlag object at 0x7f8d42b21b90>.created
E        +  and   datetime.datetime(2019, 12, 2, 14, 16, 11) = datetime(year=2019, month=12, day=2, hour=14, minute=16, second=11)
```

(You can get to logs via links commented by zuul and selecting the `Console` tab on the top of the page.)

TomasJani: Thanks for feedback. I'll start working on it right away.

ogr/abstract.py
```diff
@@ -581,7 +585,9 @@ def __str__(self) -> str:
             f"context='{self.context}', "
             f"uid='{self.uid}',"
             f"comment='{self.comment}',"
-            f"url='{self.url}')"
+            f"url='{self.url}'"
+            f"created='{self.created}'"
+            f"edited='{self.edited}')"
```
TomasJani, line 590: I should point out, that GitLab does not have <edited> property in its status object. So this may be a problem.

lachmanfrantisek: You can raise a `NotImplementedError` and add a comment with this info to the original GitHub issue.

lachmanfrantisek: To make it possible, we can have properties for those values, so you can raise that error only when someone calls that.

TomasJani: Okey, I adjusted the properties so it should be fine now.

ogr/services/gitlab/flag.py
```diff
@@ -106,3 +110,15 @@ def set(
         }
         raw_status = commit_object.statuses.create(data_dict)
         return GitlabCommitFlag(raw_commit_flag=raw_status, project=project)
+
+    @property
+    def edited(self) -> datetime.datetime:
+        # Gitlab edited not implemented
+        # https://github.com/packit-service/ogr/issues/344
+        raise NotImplementedError()
```
lachmanfrantisek, line 118: Thanks!

ogr/abstract.py
```diff
@@ -607,6 +613,22 @@ def set(
     ) -> "CommitFlag":
         raise NotImplementedError()
 
+    @property
+    def created(self) -> datetime.datetime:
+        return self._created
+
+    @created.setter
+    def created(self, new_created: datetime.datetime) -> None:
+        self._created = new_created
```
lachmanfrantisek, line 622: Does setter make sense? We probably cannot change that value.

ogr/services/github/flag.py
```diff
@@ -45,6 +45,8 @@ def _from_raw_commit_flag(self):
         self.comment = self._raw_commit_flag.description
         self.url = self._raw_commit_flag.target_url
         self.uid = self._raw_commit_flag.id
+        self.created = self._raw_commit_flag.created_at
+        self.edited = self._raw_commit_flag.updated_at
```
lachmanfrantisek, line 49: What about implementing this in the property as well so it's loaded on-demand when called?

It can look like this:
- abstract -> property with `raise NotImplementedError`
- github -> property with `return self._raw_commit_flag.created_at`
- pagure -> same as github, property using `_raw_commit_flag`
- gitlab -> Nothing, or raise `NotImplementedError` with the linked issue

TomasJani:
Your approach makes a lot more sense so I updated my version. I hope I have correctly understood your requirements.

PR #378, opened 2020-04-14
use rpmautospec

TomasTomecek:
https://docs.pagure.org/Fedora-Infra.rpmautospec/principle.html


fedora/changelog
```diff
@@ -1,53 +1,3 @@
...
-%files -n python3-%{srcname}
-%license LICENSE
-%doc README.md
-%{python3_sitelib}/%{srcname}
-%{python3_sitelib}/%{srcname}-%{version}-py?.?.egg-info
-
-
-%changelog
 * Wed Apr 01 2020 Jan Sakalos <sakalosj@gmail.com> - 0.11.1-1
 - patch release: 0.11.1
```
lachmanfrantisek, line 2: Do we need this file? Cannot it use only git log?

lachmanfrantisek:
recheck

fedora/changelog
TomasTomecek: rpmautospec doesn't process git log entriess past the latest build.

You are asking here to drop all the changelog entries which is fine to do in upstream, but terrible to do in Fedora. Hence we could not include the file here and have it downstream only, which makes sense to me.

TomasTomecek:
> Looks good. You probably want to release soon, don't you?

no need to, I'm fine to wait.

fedora/changelog
lachmanfrantisek: OK, I get it now. I misunderstood that part from their docs.

fedora/python-ogr.spec
```diff
@@ -0,0 +1,51 @@
...
+%package -n     python3-%{srcname}
+Summary:        %{summary}
+%{?python_provide:%python_provide python3-%{srcname}}
+
+%description -n python3-%{srcname}
+One Git library to Rule!
+
+
+%prep
+%autosetup -n ogr-0.11.1
```
jpopelka, line 30: Sorry for being late for the party, but why can't we use `%{srcname}-%{version}`?
I don't see it mentioned in https://docs.pagure.org/Fedora-Infra.rpmautospec/principle.html

TomasTomecek: because I'm an ignorant moron, obviously https://github.com/packit-service/ogr/pull/388

PR #381, opened 2020-04-15
Set PR flags in Pagure

csomh:
Flags on PRs are specific to Pagure.

Other forges allow setting flags/statuses only on commits and will
display the flags for the PR head on the PR page, if any is set.

Due to the above this method is added only to PagurePullRequests, and we
expect OGR library users to implement the "show the flags of the last
commit for the PR" behaviour for Pagure, in case they need it.

### DIFF
--- a/ogr/services/pagure/pull_request.py
+++ b/ogr/services/pagure/pull_request.py
@@ -22,9 +22,9 @@
 
 import datetime
 import logging
-from typing import List, Optional, Dict, Any
+from typing import List, Optional, Dict, Any, Union
 
-from ogr.abstract import PRStatus, CommitFlag
+from ogr.abstract import PRStatus, CommitFlag, CommitStatus
 from ogr.abstract import PullRequest, PRComment
 from ogr.exceptions import PagureAPIException
 from ogr.services import pagure as ogr_pagure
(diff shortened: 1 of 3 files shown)


ogr/services/pagure/pull_request.py
```diff
@@ -235,3 +235,25 @@ def merge(self) -> "PullRequest":
...
+
+    def set_flag(
+        self,
+        username: str,
+        comment: str,
+        url: str,
+        status: Optional[CommitStatus] = None,
+        percent: Optional[int] = None,
+        uid: Optional[str] = None,
+    ) -> dict:
```
lachmanfrantisek, line 247: Maybe just a docstring?

csomh: Added.

ogr/services/pagure/pull_request.py
```diff
@@ -235,3 +235,43 @@ def merge(self) -> "PullRequest":
...
+        for a full description of the parameters.
+
+        :param username: The name of the application to be presented to users
+                         on the pull request page.
+        :param comment: A short message summarizing the presented results.
+        :param url: A URL to the result of this flag.
+        :param status: The status of the task.
+        :param percent: A percentage of completion compared to the goal.
+        :param uid: A unique identifier used to identify a flag on a pull-request.
+        :return: dict with the response received from Pagure.
```
lachmanfrantisek, line 264: What about some class for that? (Or at least `CommitFlag` with some information not set?)

But we can create an issue and leave that for the future.

ogr/services/pagure/pull_request.py
```diff
@@ -235,3 +235,43 @@ def merge(self) -> "PullRequest":
...
+
+            https://pagure.io/api/0/#pull_requests-tab
+
+        for a full description of the parameters.
+
+        :param username: The name of the application to be presented to users
+                         on the pull request page.
+        :param comment: A short message summarizing the presented results.
+        :param url: A URL to the result of this flag.
+        :param status: The status of the task.
```
lachmanfrantisek, line 261: Some nitpicking: I would mention the `CommitStatus` enum or some example. `task` and `status` are unfortunately very general terms.

csomh: Removed "task" and made it more verbose.

I wouldn't add any reference to the type here, as it would duplicate the type annotation from the signature.

ogr/services/pagure/pull_request.py
csomh: Created #384.

PR #382, opened 2020-04-15
packit.xml: set upstream_package_name

TomasTomecek:
we are setting this so we can use packit from ogr's dist-git
packit can't know what's the upstream name when running from distgit

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -5,6 +5,9 @@ synced_files:
 # https://packit.dev/docs/configuration/#top-level-keys
 downstream_package_name: python-ogr
 upstream_project_url: https://github.com/packit-service/ogr
+# we are setting this so we can use packit from ogr's dist-git
+# packit can't know what's the upstream name when running from distgit
+upstream_package_name: ogr
 actions:
   # we need this b/c `git archive` doesn't put all the metadata in the tarball:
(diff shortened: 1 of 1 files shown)


TomasTomecek:
recheck

.packit.yaml
```diff
@@ -5,6 +5,9 @@ synced_files:
 # https://packit.dev/docs/configuration/#top-level-keys
 downstream_package_name: python-ogr
 upstream_project_url: https://github.com/packit-service/ogr
+# we are setting this so we can use packit from ogr's dist-git
+# packit can't know what's the upstream name when running from distgit
```
lachmanfrantisek, line 9: What trying repo name?

TomasTomecek: it was failing on the fact that tarball name started as python-ogr, instead of ogr - when I ran commands from dist-git

lachmanfrantisek: That was just an idea of what "magic" we can do in packit. (But by default, we pick the downstream name so it will collide with this approach.)

csomh:
**X**ml? o_O

TomasTomecek:
back in the day we were using [anymarkup](https://github.com/bkabrda/anymarkup) so you could write packit config in xml :D :D I always like joking about this

PR #395, opened 2020-04-24
Add is_fork and username to PagureProject __str__

mfocko:


mfocko:
@lachmanfrantisek Could you review and let me know if it solves you problem?

PR #397, opened 2020-04-25
Switch backticks to apostrophes in Pagure errors

### LINKED ISSUE #396: Use apostrophe rather than backtics in log messages.
As can be seen in https://github.com/abrt/reportd/issues/10 both [ogr](https://github.com/packit-service/ogr/blob/master/ogr/services/pagure/service.py#L150) and [packit-service](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/handlers/github_handlers.py#L179) use backticks, which results in strange Github comments.

Can ogr use apostrophes instead?

mfocko:
Fixes #396

### DIFF
--- a/ogr/services/pagure/service.py
+++ b/ogr/services/pagure/service.py
@@ -134,7 +134,7 @@ class PagureService(BaseGitService):
                 else None
             )
             raise PagureAPIException(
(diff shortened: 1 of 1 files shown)


mfocko:
@jpopelka Could you review please?
I'm going to check some API calls with apostrophes/quotes.

PR #399, opened 2020-04-27
spec: don't do python_provide on F33+

TomasTomecek:
packaging guidelines say this is automatic since F33 and we should
disable it

https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_provides

### DIFF
--- a/fedora/python-ogr.spec
+++ b/fedora/python-ogr.spec
@@ -20,7 +20,11 @@ One Git library to Rule!
 
 %package -n     python3-%{srcname}
 Summary:        %{summary}
+
+# https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_provides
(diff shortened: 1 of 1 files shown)


fedora/python-ogr.spec
```diff
@@ -20,7 +20,11 @@ One Git library to Rule!
 
 %package -n     python3-%{srcname}
 Summary:        %{summary}
+
+# https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_provides
+%if 0%{?fedora} >= 33
```
jpopelka, line 25:
```suggestion
%if 0%{?fedora} < 33
```

TomasTomecek: I'm so glad we're doing those reviews

PR #401, opened 2020-04-27
Enhance project references in pull requests

### LINKED ISSUE #400: source_project property for PR class
Now, we have only `project` property in the PR class, but it's hard to get the source project. (It can be sometimes very tricky to pick the right repo from the OGR's user perspective.)

- [ ] add `source_project` property (abstract and implementations) returning the project with the source branch
- [ ] create tests:
  - [ ] basic PR from `fork` to `upstream`
  - [ ] renamed `fork`
  - [ ] renamed `upstream`
  - [ ] PR from `fork` to `fork`
  - [ ] PR from `upstream` to `upstream`
- [ ] ? `target_project` property that will replace the current `project` one to make it clear
  - [ ] don't forget to support both and deprecate the old one

mfocko:
Closes #400

# TODO

(copied from issue)

- [x] add `source_project` property (abstract and implementations) returning the project with the source branch
- [x] create tests:
  - [x] basic PR from `fork` to `upstream`
  - [x] renamed `fork`
  - [x] renamed `upstream`
  - [x] PR from `fork` to `fork`
  - [x] PR from `upstream` to `upstream`
- [x] ? `target_project` property that will replace the current `project` one to make it clear
  - [x] don't forget to support both and deprecate the old one
-  [x] squash commits
-  [x] check getting projects by internal representation (github: creating from github.Repository, gitlab: from project_id)
   -  [x] Github: project uses almost the same way of initializing `GithubProject` in `L129, L172, L385, L564`

# Discussion

-  [x] deprecating `.project`
   -  [x] provide `.project` as a property?
   -  [ ] if so, do we need setters for them outside class?
-  [ ] consider having `GithubProject.from_github_repository`, which would take just the github repo from PyGithub and return new project
-  [ ] consider creating GitLab projects just from _project id_ (python-gitlab doesn't offer any way to get to source project than ID)

### DIFF
--- a/ogr/services/github/pull_request.py
+++ b/ogr/services/github/pull_request.py
@@ -105,15 +105,10 @@ class GithubPullRequest(BasePullRequest):
     @property
     def source_project(self) -> "ogr_github.GithubProject":
         if self._source_project is None:
-            src_github_repo = self._raw_pr.head.repo
-
-            self._source_project = ogr_github.GithubProject(
-                repo=src_github_repo.name,
-                service=self._target_project.service,
-                namespace=src_github_repo.owner.login,
-                github_repo=src_github_repo,
-                read_only=self._target_project.service.read_only,
+            self._source_project = self._target_project.service.get_project_from_github_repository(
+                self._raw_pr.head.repo
             )
+
         return self._source_project
 
     def __str__(self) -> str:
--- a/ogr/services/github/pull_request.py
+++ b/ogr/services/github/pull_request.py
@@ -39,7 +39,8 @@ logger = logging.getLogger(__name__)
 
 class GithubPullRequest(BasePullRequest):
     _raw_pr: _GithubPullRequest
-    project: "ogr_github.GithubProject"
+    _target_project: "ogr_github.GithubProject"
+    _source_project: "ogr_github.GithubProject" = None
 
     @property
     def title(self) -> str:
@@ -101,8 +102,19 @@ class GithubPullRequest(BasePullRequest):
     def head_commit(self) -> str:
         return self._raw_pr.head.sha
 
+    @property
     def source_project(self) -> "ogr_github.GithubProject":
-        raise NotImplementedError()
+        if self._source_project is None:
+            src_github_repo = self._raw_pr.head.repo
+
+            self._source_project = ogr_github.GithubProject(
+                repo=src_github_repo.name,
+                service=self._target_project.service,
+                namespace=src_github_repo.owner.login,
+                github_repo=src_github_repo,
+                read_only=self._target_project.service.read_only,
+            )
+        return self._source_project
 
     def __str__(self) -> str:
         return "Github" + super().__str__()
@@ -190,7 +202,7 @@ class GithubPullRequest(BasePullRequest):
         if not any([commit, filename, row]):
             comment = self._raw_pr.create_issue_comment(body)
         else:
-            github_commit = self.project.github_repo.get_commit(commit)
+            github_commit = self._target_project.github_repo.get_commit(commit)
             comment = self._raw_pr.create_comment(body, github_commit, filename, row)
         return GithubPRComment(parent=self, raw_comment=comment)
 
--- a/ogr/services/github/pull_request.py
+++ b/ogr/services/github/pull_request.py
@@ -101,6 +101,9 @@ class GithubPullRequest(BasePullRequest):
     def head_commit(self) -> str:
         return self._raw_pr.head.sha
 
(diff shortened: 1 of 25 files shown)


lachmanfrantisek:
>  -  [ ] deprecating `.project`
>     -  [ ] provide `.project` as a property?

Yes, probably the only way how to raise a deprecated warning.

>    -  [ ] if so, do we need setters for them outside class?

It does not make sense to have a setter for PR source/target project. You cannot change it.

> -  [ ] consider having `GithubProject.from_github_repository`, which would take just the github repo from PyGithub and return new project

If it is some implementation detail than we can create it.
- What is the use-case?
- Where do we get the service instance? (Maybe add it as a method of the `GithubService`?)

> -  [ ] consider creating GitLab projects just from _project id_ (python-gitlab doesn't offer any way to get to source project than ID)

Same as above.

mfocko:
> It does not make sense to have a setter for PR source/target project. You cannot change it.

perfect

> * What is the use-case?
> * Where do we get the service instance? (Maybe add it as a method of the `GithubService`?)

I've seen similar piece of code regarding `parent` and forking in `*Project`, so we could get rid of code in few places...

I think it could be added as `*Service` method.

mfocko:
@lachmanfrantisek not sure what you meant by renamed upstream :P could you explain?

Also regarding pagure:
-  cannot rename fork
-  cannot create PR from fork to fork
-  cannot create PR from one fork to another fork

lachmanfrantisek:
> @lachmanfrantisek not sure what you meant by renamed upstream :P could you explain?

e.g. you fork a project and then, someone renames the project upstream project => we can't use `{user}/{current_project_name}` to get the fork since fork is not renamed with the upstream project

(or vice versa, you can rename the fork)

ogr/services/gitlab/pull_request.py
```diff
@@ -93,6 +94,15 @@ def labels(self) -> List[str]:
...
         return f"{self._raw_pr.web_url}/diffs"
 
+    @property
+    def source_project(self) -> "ogr_gitlab.GitlabProject":
+        if self._source_project is None:
+            source_project_id = self._raw_pr.attributes["source_project_id"]
+            self._source_project = self._target_project.service.get_project(
+                iid=source_project_id
+            )
+        return self._source_project
```
lachmanfrantisek, line 104: Is it possible to use the same approach as you use for GitHub and avoid changing the `get_project` method?

mfocko:
> I would only try to use a similar approach for all implementations. You use a slightly different way for each of them.
> 
> Is it possible to use `Pagure` or `Github` way for all of them?

That's exactly what I was intending to look into after the tests. The problem is basically that Github and Pagure store the project right away in the PR (github as its own object (specific to library), Pagure as dictionary), so I thought about method for objects created from library-specific objects?

However the GitLab is different story, they (apparently) keep only project ID in PR.

Regarding GitHub, I've found 4 places in `GithubProject` that use the similar way to create project as I do in `source_project` for PR, so it could be factored out, but still can't make the interface unified. (gitlab has it in 2 places I think, not so similar to each other though)

lachmanfrantisek:
> The problem is basically that Github and Pagure store the project right away in the PR (github as its own object (specific to library), Pagure as dictionary), so I thought about method for projects created from library-specific objects?

I am not against, I just don't like to have this in the api -- do this internally, in the `_methods`, or `__methods`...

ogr/services/github/service.py
```diff
@@ -142,6 +142,17 @@ def get_project(
             **kwargs,
         )
 
+    def get_project_from_github_repository(
+        self, github_repo: github.Repository.Repository
+    ) -> "GithubProject":
```
lachmanfrantisek, line 147: :+1:

ogr/services/github/pull_request.py
```diff
@@ -97,6 +98,20 @@ def labels(self) -> List[GithubLabel]:
...
+        if self._source_project is None:
+            src_github_repo = self._raw_pr.head.repo
+
+            self._source_project = ogr_github.GithubProject(
+                repo=src_github_repo.name,
+                service=self._target_project.service,
+                namespace=src_github_repo.owner.login,
+                github_repo=src_github_repo,
+                read_only=self._target_project.service.read_only,
+            )
```
lachmanfrantisek, line 112: Maybe I am missing something, but cannot we use the `service.get_project_from_github_repository`?

mfocko: Thanks, forgot it in the place it was made for :D

mfocko:
@lachmanfrantisek Could you have a look please and let me know if it resolves your problem?

PR #415, opened 2020-05-22
Add PullRequest.patch property

jpopelka:
Currently implemented only for PagurePullRequest.

It returns bytes in case there's some binary blob in the patch.

I'd like to use this in https://github.com/packit-service/packit-service/pull/627

### DIFF
--- a/ogr/services/pagure/pull_request.py
+++ b/ogr/services/pagure/pull_request.py
@@ -106,6 +106,17 @@ class PagurePullRequest(BasePullRequest):
         return f"{self.url}#request_diff"
 
     @property
+    def patch(self) -> bytes:
+        request_response = self._target_project._call_project_api_raw(
+            "pull-request", f"{self.id}.patch", add_api_endpoint_part=False
+        )
+        if request_response.status_code != 200:
+            raise PagureAPIException(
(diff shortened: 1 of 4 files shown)


ogr/services/pagure/pull_request.py
```diff
@@ -105,6 +107,14 @@ def created(self) -> datetime.datetime:
     def diff_url(self) -> str:
         return f"{self.url}#request_diff"
 
+    @property
+    def patch(self) -> bytes:
+        patch_url = f"{self.url}.patch"
+        response = get(patch_url)
```
lachmanfrantisek, line 113: I am only if we can use the session used across the whole codebase. (`self.__call_api`)

mfocko: `self.__call_api` automatically uses API endpoint, also `patch_url` points to raw file not JSON
maybe  `_call_project_api_raw`?

jpopelka: Updated, can you check again?

jpopelka:
Can you give final thumbs up?

> Can you please create an issue for other implementation after the merge?

https://github.com/packit-service/ogr/issues/416

lachmanfrantisek:
> Can you give final thumbs up?

Sorry @jpopelka, I missed that.

PR #419, opened 2020-05-26
Pull requests on Gitlab

### LINKED ISSUE #414: Pull requests on Gitlab projects
By looking at the code I noticed there are no tests for creating pull requests in different scenarios (like Github has) and also the code doesn't seem to be compliant to the documentation of ogr.

-  [ ] Add tests
-  [ ] Fix the implementation

May be blocked by #412

mfocko:
Closes #414

### DIFF
--- a/ogr/services/gitlab/pull_request.py
+++ b/ogr/services/gitlab/pull_request.py
@@ -26,6 +26,7 @@ from typing import List, Optional
 from gitlab.v4.objects import MergeRequest as _GitlabMergeRequest
 
 from ogr.abstract import PullRequest, PRComment, PRStatus
+from ogr.exceptions import GitlabAPIException
 from ogr.services import gitlab as ogr_gitlab
 from ogr.services.base import BasePullRequest
 from ogr.services.gitlab.comments import GitlabPRComment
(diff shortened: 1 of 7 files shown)


ogr/services/gitlab/pull_request.py
```diff
@@ -118,16 +119,50 @@ def create(
...
+            other_project = GitlabPullRequest.__get_fork(
+                fork_username,
+                project.get_forks()
+                if project.parent is None
+                else project.parent.get_forks(),
+            )
+
+            parameters["target_project_id"] = other_project.gitlab_repo.attributes["id"]
+            if project.parent is None:
+                parameters["target_project_id"] = repo.attributes["id"]
```
mfocko, line 146: maybe use one variable `target_id: Optional[int]` and just assign it if has value at the end? Would be cleaner

ogr/services/gitlab/pull_request.py
```diff
@@ -118,16 +119,50 @@ def create(
...
+                repo = other_project.gitlab_repo
+
+        mr = repo.mergerequests.create(parameters)
         return GitlabPullRequest(mr, project)
 
+    @staticmethod
+    def __get_fork(
+        fork_username: str, forks: List["ogr_gitlab.GitlabProject"]
```
mfocko, line 154: just pass the project in to move `get_forks` call here?

ogr/services/gitlab/pull_request.py
```diff
@@ -118,16 +119,50 @@ def create(
...
+            "description": body,
+        }
+
+        if project.is_fork and (
+            fork_username is None or fork_username == project.namespace
+        ):
+            parameters["target_project_id"] = project.parent.gitlab_repo.attributes[
+                "id"
+            ]
+        elif fork_username and fork_username != project.namespace:
```
lachmanfrantisek, line 136: Maybe some comment about `else` variant?

ogr/services/gitlab/pull_request.py
lachmanfrantisek: Yes, sounds good. Or, document/name this method properly. It takes a while to understand what is going on here.

(e.g. Why you search for fork and you already send a list of forks... It's more about searching for the right fork than getting the fork(s).)

ogr/services/gitlab/pull_request.py
```diff
@@ -118,16 +119,50 @@ def create(
...
+    ) -> "ogr_gitlab.GitlabProject":
+        forks = list(
+            filter(
+                lambda fork: fork.gitlab_repo.namespace["full_path"] == fork_username,
+                forks,
+            )
+        )
+        if not forks:
+            raise GitlabAPIException("Requested fork doesn't exist")
+        return forks[0]
```
lachmanfrantisek, line 164: Is there a chance that we get more than one?

tests/integration/test_gitlab.py
```diff
@@ -552,6 +552,92 @@ def test_source_project_renamed_upstream(self):
         assert source_project.namespace == "mfocko"
         assert source_project.repo == "new-ogr-testing-repo-in-the-group"
 
+    def test_create_pr_upstream_upstream(self):
```
lachmanfrantisek, line 555: Thanks for all those tests.

ogr/services/gitlab/pull_request.py
mfocko: It filters by username, so it shouldn't.
However after GitLab allowing to create MR from non-existing branch through API, I wouldn't be surprised :D 

More straight-forward would be to get the project directly, but that introduces problem with renamed project (either one).

ogr/services/gitlab/pull_request.py
mfocko: I'll add those, also usage as in github implementation

PR #425, opened 2020-06-08
Implement PagurePullRequest.get_flags()

jpopelka:
Needed in https://github.com/packit-service/packit-service/pull/662

### DIFF
--- a/ogr/services/pagure/pull_request.py
+++ b/ogr/services/pagure/pull_request.py
@@ -267,6 +267,9 @@ class PagurePullRequest(BasePullRequest):
         self.__update()
         return self.project.get_commit_statuses(self._raw_pr["commit_stop"])
 
+    def get_flags(self) -> List[Dict]:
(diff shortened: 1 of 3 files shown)


ogr/services/pagure/pull_request.py
```diff
@@ -267,6 +267,9 @@ def get_statuses(self) -> List[CommitFlag]:
         self.__update()
         return self.project.get_commit_statuses(self._raw_pr["commit_stop"])
 
+    def get_flags(self) -> List[Dict]:
+        return self.__call_api("flag", method="GET")["flags"]
```
lachmanfrantisek, line 271: Using `GET` is the default, but I am not against explicitness since it's deep in the call chain (in `PagureService.call_api_raw`).

jpopelka:
Thank you for your advices!
I don't need this after all since I'm using the commit statuses now and only set the PR flag.
Closing.

PR #436, opened 2020-07-16
Support add group for pagure

dhodovsk:


tests/integration/test_pagure.py
```diff
@@ -95,6 +95,10 @@ def test_add_user(self):
         project = self.service.get_project(repo="hello-112111", namespace="testing",)
         project.add_user("lachmanfrantisek", AccessLevel.admin)
 
+    def test_add_group(self):
+        project = self.service.get_project(repo="hello-112111", namespace="testing",)
+        project.add_group("packit-test-group", AccessLevel.admin)
```
lachmanfrantisek, line 100: You need to run this test locally and attach the response file to the PR.

Do you need some help with that? You just need to provide pagure token as an environment variable and use real values.

PR #439, opened 2020-07-26
Requesting project access

shreyaspapi:
Request project access - using it here (https://github.com/packit-service/packit-service/pull/740)

[Request access docs](https://python-gitlab.readthedocs.io/en/stable/gl_objects/access_requests.html)

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -885,7 +885,18 @@ class GitProject:
         :param access_level: AccessLevel enum
         :return: None
         """
-        raise NotImplementedError
+        raise NotImplementedError()
+
+    def request_access(self) -> None:
+        """
+        Access request will be sent to the project,
+        you cannot specifically ask for role.
+        The selection of the role and approval needs to be
+        done by the user with maintainer/admin rights.
+
+        :return: None
+        """
+        raise NotImplementedError()
 
     def add_group(self, group: str, access_level: AccessLevel) -> None:
         """
--- a/ogr/services/gitlab/project.py
+++ b/ogr/services/gitlab/project.py
@@ -241,6 +241,12 @@ class GitlabProject(BaseGitProject):
         except Exception as e:
             raise GitlabAPIException(f"User {user} already exists", e)
 
+    def request_access(self) -> None:
+        try:
+            self.gitlab_repo.accessrequests.create({})
+        except gitlab.exceptions.GitlabCreateError as e:
+            raise GitlabAPIException("Unable to request access", e)
+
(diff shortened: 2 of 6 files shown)


ogr/abstract.py
```diff
@@ -885,7 +885,15 @@ def add_user(self, user: str, access_level: AccessLevel) -> None:
         :param access_level: AccessLevel enum
         :return: None
         """
-        raise NotImplementedError
+        raise NotImplementedError()
+
+    def request_access(self) -> None:
```
lachmanfrantisek, line 890: Does it make sense to request a different role?

At least, please, document the role we are requesting.

shreyaspapi: I did consider this.

`project.accessrequests.create()` does not take any arguments. We cannot request a specific role, User while approving will have to specify a role.

lachmanfrantisek: OK, can you please add this note to the docstring?

And it would be nice to be able to provide some description/note that will be shown to the fork owner. But it looks like we can't set anything else via GitLab API (https://docs.gitlab.com/ce/api/access_requests.html).

shreyaspapi: Added in the docstring.

ogr/abstract.py
```diff
@@ -885,7 +885,17 @@ def add_user(self, user: str, access_level: AccessLevel) -> None:
...
+        """
+        Access request will be sent to the project,
+        you cannot specifically ask for role.
+        The request will only be able to approve by user with maintainer / admin rights
```
lachmanfrantisek, line 894:
```suggestion
        The selection of the role and approval needs to be done by the user with maintainer/admin rights.
```

ogr/services/gitlab/project.py
```diff
@@ -241,6 +241,12 @@ def add_user(self, user: str, access_level: AccessLevel) -> None:
         except Exception as e:
             raise GitlabAPIException(f"User {user} already exists", e)
 
+    def request_access(self) -> None:
+        try:
+            self.gitlab_repo.accessrequests.create({})
+        except Exception as e:
```
lachmanfrantisek, line 247: Can we use some Gitlab specific exception instead? (Don't we discuss this somewhere?..;)

shreyaspapi: Yes on a different repo😅. Changing it.

ogr/abstract.py
shreyaspapi: Done!

ogr/services/gitlab/project.py
shreyaspapi: Done!

PR #450, opened 2020-08-11
Use tokman in ogr

mfocko:
-  [x] Interface for token manager using github app
-  [x] Refactor default solution
-  [x] Implement support for `tokman`


ogr/services/github/service.py
```diff
@@ -177,3 +187,9 @@ def project_create(self, repo: str, namespace: str = None) -> "GithubProject":
             service=self,
             github_repo=new_repo,
         )
+
+    def get_github_instance(self, repo: str, namespace: str) -> github.Github:
```
csomh, line 191: Wouldn't it be more intuitive to have namespace first, followed by repo? :)

ogr/services/github/token_managers/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
```
csomh, line 21: How about using the short licence headers:

```
# Copyright Contributors to the Packit project.
# SPDX-License-Identifier: MIT

```

ogr/abstract.py
```diff
@@ -1394,3 +1394,20 @@ def get_projects(self) -> Sequence["GitProject"]:
 
     def get_forks(self) -> Sequence["GitProject"]:
         raise NotImplementedError
+
+
+class GithubTokenManager:
+    """
+    Represents a token manager for authentication via GitHubApp.
```
lachmanfrantisek, line 1401: I would place a note about remote service we need to run for that.

```suggestion
    Represents a token manager for authentication via GitHubApp using a remote REST service.
```

ogr/abstract.py
```diff
@@ -1394,3 +1394,20 @@ def get_projects(self) -> Sequence["GitProject"]:
...
+    """
+
+    def get_instance(self, namespace: str, repo: str) -> Any:
```
lachmanfrantisek, line 1404:
```suggestion
    def get_instance(self, namespace: str, repo: str) -> github.Github:
```

ogr/abstract.py
```diff
@@ -1394,3 +1394,20 @@ def get_projects(self) -> Sequence["GitProject"]:
...
+        """
+        Returns GitHub instance for a requested repository,
```
lachmanfrantisek, line 1406: The term `instance` is a bit misleading -- would be nice to explicitly state that we are talking about Github instance from pygithub.

ogr/services/github/service.py
```diff
@@ -76,6 +85,16 @@ def github_app_private_key(self):
...
 
+    @property
+    def token_manager(self) -> GithubTokenManager:
+        if self._token_manager is None:
+            self._token_manager = (
+                TokmanGithubTokenManager(self.tokman_instance_url)
+                if self.tokman_instance_url
+                else OgrGithubTokenManager(self)
+            )
+        return self._token_manager
```
lachmanfrantisek, line 96: What about basic token-based authentication. Don't we want to ignore those managers (return `None`?) in that case?

ogr/abstract.py
mfocko: Not necessarily, the default token manager (when you provide _GitHub App ID_ and key) doesn't use REST and utilizes `PyGithub` directly.

lachmanfrantisek: OK, now I get the architecture.

lachmanfrantisek: What about renaming the class to  `GithubAppTokenManager`?

mfocko: > What about renaming the class to `GithubAppTokenManager`?

The way `PyGithub` instances are retrieved, you could create a subclass that uses personal access token**s**. Even the default token auth that is included in `GithubService` could be extracted into `TokenManager` subclass. (That would be probably a mess to handle in `__init__` (since it's passed by raw data, e.g. token), but it would remove all the logic regarding auth from `GithubService`)

(Also the subclasses already look like Java classes because of the naming) :/

lachmanfrantisek: That was my point that you use `TokenManager` only for Github apps so it makes sense to name it like that.

Will it be so hard to handle token auth in the token manager subclass to have it consistent and exclude the authentication completely?

mfocko: The main problem is in service constructor, since it takes info of each auth method in separate arguments, that is what decides if token or token manager (either Tokman or github app through pygithub) is used and it's done in service.
(Which also makes checking for multiple provided auth methods interesting (passing tuple to check if only one is not None))

Separate classes for each auth method would provide nice encapsulation, but I wonder how much code is tied to public attributes related to that. Creating those objects probably from factory-like method, like is used for parsing services?

lachmanfrantisek: > The main problem is in service constructor

Yes, I got that but still think we can try to separate that...;)

>  I wonder how much code is tied to public attributes related to that.

I am not sure but probably mostly tests. The functionality should not use that.

> Creating those objects probably from factory-like method, like is used for parsing services?

:+1:

ogr/services/github/service.py
```diff
@@ -60,6 +66,21 @@ def __init__(
         self.github = github.Github(login_or_token=self.token)
         self.read_only = read_only
 
+        self.tokman_instance_url = tokman_instance_url
+        self._token_manager = token_manager
+
+        self.__check_for_multiple_auths(
+            (token, github_app_id, tokman_instance_url, token_manager)
+        )
```
lachmanfrantisek, line 74: What about hiding all attributes related to the authentication into the token manager(s)?

lachmanfrantisek: So have a method that will receive all auth-related attributes and return an instance of token manager?

ogr/services/github/service.py
```diff
@@ -177,3 +218,13 @@ def project_create(self, repo: str, namespace: str = None) -> "GithubProject":
...
+
+    def get_github_instance(self, namespace: str, repo: str) -> github.Github:
+        if (
+            self._token_manager
+            or self.tokman_instance_url
+            or (self.github_app_id and self.github_app_private_key)
+        ):
+            # authenticating as a GitHub app or using Tokman
+            return self.token_manager.get_pygithub_instance(namespace, repo)
+        return self.github
```
lachmanfrantisek, line 230:
```suggestion
    def get_pygithub_instance(self, namespace: str, repo: str) -> github.Github:
        if (
            self._token_manager
            or self.tokman_instance_url
            or (self.github_app_id and self.github_app_private_key)
        ):
            # authenticating as a GitHub app or using Tokman
            return self.token_manager.get_pygithub_instance(namespace, repo)
        return self.github
```

ogr/services/github/service.py
```diff
@@ -47,64 +56,53 @@ def __init__(
...
+                github_app={
+                    "id": github_app_id,
+                    "private_key": github_app_private_key,
+                    "private_key_path": github_app_private_key_path,
+                }
+                if github_app_id
+                else None,
+                tokman_instance_url={"instance_url": tokman_instance_url}
+                if tokman_instance_url
+                else None,
```
lachmanfrantisek, line 79: Does it make sense to do this magic? I am only thinking if we can directly create that instance when we have those `if`s here.

If the check for multiple methods is the problem. Define priorities in docstring and skip it...;)

lachmanfrantisek (commented):
I really like how you have moved the authentication to different classes. Nice!

I have only one concern about the construction of those objects. (Sorry...;)

ogr/services/github/token_managers/tokman_manager.py
```diff
@@ -0,0 +1,43 @@
...
+from ogr.exceptions import OgrException
+
+
+class TokmanGithubTokenManager(GithubTokenManager):
+    def __init__(self, instance_url: str):
+        self.__instance_url = instance_url
+
+        try:
+            # check if running correctly
+            health = requests.get(f"{self.__instance_url}/api/health")
```
csomh: Is doing this check here really required?

My concern is that this will create a dependency on the tokman service running at construction time, which is not really required: it should be completely fine to have the tokman service down when the object is created, the only thing that matters is to have it running when one tries to retrieve a token.

In other words, doing this check here, will make the design rigid.

Am I missing something?

ogr/services/github/token_managers/tokman_manager.py
```diff
@@ -0,0 +1,43 @@
...
+
+        json_response = response.json()
+        token = json_response.get("access_token", None)
+
+        if not response.ok or not token:
+            # haven't managed to get a token
+            error = json_response.get("error", "no message provided")
+            raise OgrException(
+                f"Couldn't retrieve token from Tokman: {error}"
+            )
```
csomh: Usually checking the response for errors is done *before* any information is retrieved from the response.

Additionally: any reason for not using `response.raise_for_status()`? Should make thing simpler.

ogr/services/github/service.py
```diff
@@ -47,64 +56,49 @@ def __init__(
...
         github_app_private_key_path: str = None,
+        tokman_instance_url: str = None,
+        github_authentication: GithubAuthentication = None,
         **_,
     ):
+        """
+        If multiple authentication methods are provided, they are prioritised:
+            1. Tokman
+            2. GithubApp
+            3. Token (which is also default one, that works without specified token)
```
csomh, line 67: I might have missed the discussion: could this be made explicit, instead of computing it, in order to remove some complexity?

ogr/services/github/auth_providers/abstract.py
```diff
@@ -0,0 +1,38 @@
...
+        authenticated using GitHub App
+
+        :param namespace: namespace of the repository
+        :param repo: name of the repository
+        :return: instance of github.Github
+        """
+        raise NotImplementedError()
+
+    @property
+    def pygithub_instance(self) -> "github.Github":
```
csomh, line 26: This design (the authentication provider returning the instance of the service where it's being authenticated) is wired. Usually it's something like:

```
service = Service(AuthProvider())
```

and then in `Service.__init__()`:

```
token = auth_provider.get_token()
```

That is: the service receives an authentication provider which then is used to get the token.

Would it make sense to follow this pattern in this case, too?

ogr/services/github/auth_providers/tokman.py
```diff
@@ -0,0 +1,64 @@
...
+        self.__healthy = False
+
+    def __eq__(self, o: object) -> bool:
+        if not issubclass(o.__class__, Tokman):
+            return False
+
+        return self._instance_url == o._instance_url  # type: ignore
+
+    def __str__(self) -> str:
+        # TODO: Check if needs to be censored
```
csomh, line 25: No, there is no need for that.

ogr/services/github/service.py
lachmanfrantisek: We are loading the services from config (yaml) where you can easily set those arguments. (The authenticator instance is possible as well.)

ogr/services/github/token_managers/tokman_manager.py
mfocko: > Am I missing something?

Nope, GitHub didn't show you latest version apparently or you reviewed commit by commit. Latest version does this check on the first call of `get_pygithub_instance`

ogr/services/github/auth_providers/abstract.py
lachmanfrantisek: Two points that might answer your question, but TBH, I don't fully understand your question/proposal:

- We are instantiating the service when loading config. Sometimes they are not used => we want to ask for the token only when really needed.
- The token is per-project for GitHub apps and does not make sense to have one for the service.

mfocko: done

(can't be moved to `__init__` because of GitHub app tokens though)

mfocko: @lachmanfrantisek it removes the construction of `github.Github` from auth objects and constructs them in service with the provided token (which basically gets rid of the "same" `return` expression in all of them)

ogr/services/github/token_managers/tokman_manager.py
mfocko: > Additionally: any reason for not using response.raise_for_status()? Should make thing simpler.

Also done, but it raises generic `BAD REQUEST` exception right now... It's in separate commit in case we would prefer raising `OgrException` with "prettier" message.

ogr/services/github/token_managers/tokman_manager.py
csomh: Yeah, true, I was checking an older commit.

Still, I hold up this comment as valid: the primary use of the `health` API endpoint was to provide a way for OpenShift to check the health of tokman.

I don't see what additional information or benefit do we get here by checking the health before requesting a token.  Whatever error code is found in the `response` should be enough to identify all issues, including the case when the tokman service is not running.

ogr/services/github/auth_providers/abstract.py
csomh: > * The token is per-project for GitHub apps and does not make sense to have one for the service.

Apologies, I always forget that this is the case :|

> @lachmanfrantisek it removes the construction of github.Github from auth objects and constructs them in service with the provided token (which basically gets rid of the "same" return expression in all of them)

Exactly this was the design I was looking for, though failed to formulate it correctly. Thanks!

ogr/services/github/token_managers/tokman_manager.py
csomh: Indeed, this will not show the error message, which is not nice.

The main issue with the previous solution is that it's using `response.json()` before checking the status codes. Although the tokman API will return JSON for 200 and 400, the response body for other status codes might not be valid JSON, in which case this call will fail.

How about not reading the response as a JSON in case of an error, but simply using whatever there is in `response.text`?

ogr/services/github/service.py
lachmanfrantisek, line 67: Post-rename leftover:

```suggestion
            3. TokenAuthentication (which is also default one, that works without specified token)
```

lachmanfrantisek (commented):
No[*] objections from my side..;) Good job!

Do you want to implement `tokman` here?

\* sorry, just one typo..;(

ogr/services/github/token_managers/tokman_manager.py
mfocko: Thanks, removed. It was intended as a way to catch problem like typos immediately after running (e.g. service), but since I wanted to enforce laziness, I reorganized it and left it there...

ogr/services/github/token_managers/tokman_manager.py
mfocko: Redone with `response.text`

ogr/services/github/service.py
mfocko: I'm not sure if there is *better* solution than either of those...

- So far it has been handled in `GithubService` silently prioritizing GitHub App
  _coughing_ even with side effect that defaulted to token if no key given and empty file provided _coughing_ (that also presented an interesting issue with tests failing for invalid key format in file...)
- If it is handled explicitly at initialization, seems like a mess too
- Current implementation allows keeping the logic inside those authentication classes with a little cost in service
---
- Checking if provided *exactly* one way to authenticate raises some complications because of the 3 attributes for GitHub App
  which then results in _creative_ dictionary that is used to resolve it later or explicit if-statements in service class...
- We could deprecate current interface and enforce passing authentication object in parameter, however that would mean problem in every place the GithubService is constructed and also (for the purpose of deprecation) would still keep the mess inside the service class
  - we would have to take care of this in `ogr/factory.py`, so it seems more like moving problem from one place to another for me?
  - if created from one of the factory functions, it shouldn't be that big of a deal, cause it could be fixed there; however explicitly using `GithubService` constructor...

csomh: I see, thanks for the explanation!

Given all this, I think it's fine to leave it as it is for now.

csomh (approved):
:rocket:

lachmanfrantisek (approved):
:tada: Nicely done! :tada: 

Thanks!

Issue #295, opened 2019-12-05
Restructure tests


mfocko:
Split tests into files like the implementations are.

As @lachmanfrantisek suggested, it would be great to have both old tests (checking that deprecated interfaces are still usable) and new ones, but this would lead to keeping duplicates of test_data, since they would check the same things, but each would need separate yaml for responses.

So I suggest transition from old to new tests right after 0.13.0 (or before 1.0.0, in case it will be released sooner)

lachmanfrantisek:
@mfocko what do you think? Do we still want it.

mfocko:
@lachmanfrantisek Yes, it's just waiting for the 0.13.0 release, since we've decided on keeping old tests to make sure the deprecated (but still present) functionality is working properly and restructuring those tests would mean keeping duplicate response files just because of different test names (and probably different call stack too).

PR #452, opened 2020-08-13
this is first draft how new tests could look like

jscotka:


tests/integration/test_factory.py
```diff
@@ -1,37 +1,31 @@
...
-            if not self.github_token:
-                raise EnvironmentError(
-                    "You are in requre write mode, please set GITHUB_TOKEN env variables"
-                )
-            if not self.pagure_token:
-                raise EnvironmentError(
-                    "You are in requre write mode, please set PAGURE_TOKEN env variables"
-                )
-            if not os.environ.get("GITLAB_TOKEN"):
+        if not os.path.exists(get_datafile_filename(obj=self)):
```
lachmanfrantisek, line 17: What about returning path from this function?

tests/integration/test_factory.py
```diff
@@ -43,7 +37,9 @@ def setUp(self):
             self.gitlab_service,
         ]
 
+    @record_requests()
     def test_get_project_github(self):
+        self.init_services()
```
lachmanfrantisek, line 42: Can we move this to the `SetUp` method so we don't need to put this into all test methods?

jscotka:
> So we can clear the `tests/__init__.py` once we move everything to this format?
right, or better to say, we have to, because now it causes confusion to requre, because there are double storing one via init and second via decorators.

tests/integration/test_factory.py
jscotka: > Can we move this to the `SetUp` method so we don't need to put this into all test methods?

I wanted that as well, but found it is impossible, because decorators are just around test functions, and setUp is done before tests executions. so services initialization is out, and when running without tokens it led to missing authorizations for setup. I wanted to decorate whole class, but it is impossible in case unittests, because tests are then not found  from some reason. maybe something connected that unittest classes are not intantiated, or something similar. 

This part is little bit tricky and previous solution, was more elegant for this, because it changed modules before anything in tests. But it makes sense that it works how it works :-)

Do you any suggestion how to deal with that? I cannot also decorate setup, because it changes cassette, so it will not be then part of cassette of test.

tests/integration/test_factory.py
```diff
@@ -1,48 +1,62 @@
...
+                    " GITLAB_TOKEN env variables"
                 )
 
-        self.github_service = GithubService(token=self.github_token)
-        self.pagure_service = PagureService(token=self.pagure_token)
-        self.gitlab_service = GitlabService(
-            token=self.gitlab_token, instance_url="https://gitlab.com"
-        )
-        self.custom_instances = [
+    @property
```
lachmanfrantisek, line 31: Is it regenerated for each test case or only once?

tests/integration/test_factory.py
lachmanfrantisek: I mean `Path` from `pathlib`.

lachmanfrantisek (commented):
Looks good.

Would be nice to have a way to decorate a class. What about adding some `@record_request_on_all_methods` class decorator to requre?

We can get inspired by this:
https://stackoverflow.com/a/6307868

(I am not sure if this will not break the response handling.)

tests/integration/test_factory.py
jscotka: for each test case , I've tried to test it, that I've regenerated all outputs and then try to run second test, and seems to work as expected.

jscotka: Theoretically also code like: 
```
    @pytest.fixture(autouse=True)
    @record_request
    def prepare_flag(self):
           yield
   ```
coud work as expected, to decorate each test method, as you've sent me link to https://stackoverflow.com/questions/50710028/how-to-use-pytest-fixtures-with-unittest-methods
but not sure if it is readable enough, and does not cause confusion,
I try to look at it and try to find the best way how to do it.

Comment (jscotka) [reply tests/integration/test_factory.py:31]: yep, I've tested, the code like: 

```
def for_all_methods(decorator, methond_pattern="test.*"):
    def decorate(cls):
        for attr in cls.__dict__:
            if callable(getattr(cls, attr)) and re.match(methond_pattern, attr):
                setattr(cls, attr, decorator(getattr(cls, attr)))
        return cls
    return decorate

@for_all_methods(record_requests)
class TestClass(unittest.TestCase)
...
```
works as expected. I can incorporate it into `requre` as well, that's true that decoration every test function, especially inside more extensive testsuite like github, will contain lots of this same decorators. what do you think?

Comment (lachmanfrantisek) [reply tests/integration/test_factory.py:31]: Nice! Is it possible to have this in some supper class?

Comment (jscotka) [reply tests/integration/test_factory.py:31]: > Nice! Is it possible to have this in some supper class?

@lachmanfrantisek not sure. therein text what you've linked is used metaclass, to override `__new__` constructor to apply these decorators, but from my perspective it will be more cryptic to use these metaclasses.

Comment (lachmanfrantisek) [reply tests/integration/test_factory.py:31]: @jscotka I would go with the class decorator and users can decide if they want to mark class or method.

What about this:

```python
def record_requests_for_all_methods(methond_pattern="*"):
    def decorate(cls):
        for attr in cls.__dict__:
            if callable(getattr(cls, attr)) and re.match(methond_pattern, attr):
                setattr(cls, attr, record_requests(getattr(cls, attr)))
        return cls
    return decorate
```

jscotka: moved generic code to https://github.com/packit/requre/pull/112 and then I'll use it here. Your suggestion will be useful just for `ogr` where is used just requests module to be decorated.

tests/integration/test_github.py
```diff
@@ -776,11 +789,13 @@ def test_source_project_other_fork_fork(self):
         assert source_project.repo == "hello-world"
 
     def test_source_project_renamed_fork(self):
+        # TODO: why there were "bye-world" in source_project.repo, it needs some manual change?
```
lachmanfrantisek, line 792: As the name of the tests says, we want to check that we get right name even if the upstream and for project name differ. E.g. when you:
- rename the upstream after the fork was forked
- rename the fork

tests/integration/test_github.py
```diff
@@ -837,13 +853,15 @@ def test_edit_release(self):
         assert release.body == f"{origin_message}-changed"
 
     def test_latest_release(self):
-        last_version = "0.7.0"
+        # check lastest release  og ogr and fix version and string in body
```
lachmanfrantisek, line 856:
```suggestion
        # check the latest release of OGR and fix version and string in the body
```

tests/integration/test_github_readonly.py
```diff
@@ -1,24 +1,34 @@
...
 
-
-class ReadOnly(RequreTestCase):
+@record_requests_for_all_methods()
+class ReadOnly(unittest.TestCase):
     def setUp(self):
         super().setUp()
         self.token = os.environ.get("GITHUB_TOKEN")
-        if PersistentObjectStorage().mode == StorageMode.write and not self.token:
+        if not Path(get_datafile_filename(obj=self)).exists() and not self.token:
```
lachmanfrantisek, line 14: Do we have the test function accessible here? What is the `self` here?

It would be nice to have `cassette` somehow accessible and we now save it as the `._requre_cassette` attribute of the function so this might be the way.

tests/integration/test_github.py
jscotka: yep, that's why I've added there this TODO, because It probably needs some manual change, or something like that and it is not described, what to do, before this test ( for others I skipped these tests.)

tests/integration/test_github_readonly.py
jscotka: self is instance of tests and I uses it as internal knowledge that this object it used for identifying test file if it is pytest, so that it is little bit cryptic.
Without this singleton it raised this issue, that in setup I cannot access cassette, theoretical solution is, similar to moving self.service to property, and just handle mode inside test, not inside setup, then I could use external cassette and look into cassette inside test part and then use your solution via `._requre_cassette` but it is not possible inside setup anyhow.
theoretically there are several ways how to deal with it:
 * do not raise exceptions and just run tests depends on setup tokens, so that then I do not need to deal with some cryptic internal get file names  function
 * your solution, but use it inside test, not setup (because cassette name is not know in setup part)
 * solution with one external cassette and just setup the cassette name inside setup and in test handle itself
 * use the singleton, and some part class for tests what will solve it.

I do not other idea how  we can solve it.

lachmanfrantisek: > * do not raise exceptions and just run tests depends on setup tokens, so that then I do not need to deal with some cryptic internal get file names  function

Last options I think.

> * your solution, but use it inside test, not setup (because cassette name is not know in setup part)

Doable, but I would try to avoid having the same call in all methods.

> * solution with one external cassette and just setup the cassette name inside setup and in test handle itself
> * use the singleton, and some part class for tests what will solve it.

I would try to avoid having global states completely.

----

It's not so easy to solve. I'll think about that but we can leave it as-is for now.

jscotka: theoretically the easiest way could be, to have that external cassette inside decorators, and in class setUp()
has code like:
```
cassette.storage_file(obj=self)
```
and then have decorator with this `cassette=cassette`

then cassette will not be setup as part of the init, because it exists, so  the code will be just part of setup, but bad is that you then have to cassette=cassette and probably could lead to some sideeffects caused by this shared cassette, similar to singleton usages.

mfocko:
Closes #295

jscotka:
@mfocko could we merge it, or we're waiting for some changes?

mfocko:
> @mfocko could we merge it, or we're waiting for some changes?

@jscotka there were some minor changes because of the tests (deprecation warnings, etc.), so I'll ping @lachmanfrantisek on Monday for review

I will also rebase and squash afterwards

lachmanfrantisek:
recheck

lachmanfrantisek (approved):
:rocket: Good job @jscotka and @mfocko!

PR #460, opened 2020-09-02
Update pre-commit configuration and fix mypy remarks

mfocko:
*  Update pre-commit
*  Fix imports from PyGithub
*  Type-cast `release.created_at` from datetime to string

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,17 +4,16 @@
 
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 20.8b1
     hooks:
       - id: black
-        language_version: python3.6
   - repo: https://github.com/prettier/prettier
-    rev: 2.0.5
+    rev: 2.1.1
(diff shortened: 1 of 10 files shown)


mfocko:
@lachmanfrantisek Could you have a look, please?

Regarding the `PyGithub` backward-compatibility for versions from RPM and pypi, I cannot seem to fix it so far:
```
ogr/services/github/auth_providers/github_app.py:85: error: "_ValuedAttribute" has no attribute "value"
ogr/services/github/auth_providers/github_app.py:92: error: Argument 1 to "get_access_token" of "GithubIntegration" has incompatible type "_ValuedAttribute"; expected "int"
```

TomasTomecek (approved):

jpopelka (approved):

lachmanfrantisek (approved):
Maybe, you can use `hasattr`/`getattr` to fix that but feel free ignore those lines (`# type: ignore`) if the old version causes troubles...

PR #463, opened 2020-09-14
Support multi part namespaces

lachmanfrantisek:
- Support multi-part namespaces in parsing.

### DIFF
--- a/ogr/parsing.py
+++ b/ogr/parsing.py
@@ -136,16 +136,22 @@ def parse_git_repo(potential_url: str) -> Optional[RepoUrl]:
         )
     if not username and len(split) >= 2:
         # path contains username/reponame
+        # or some/namespace/reponame
+        # or fork/username/some/namespace/reponame
 
-        is_fork = "fork" in split
-        if is_fork and len(split) == 4:
+        is_pagure_fork = "fork" == split[0] and len(split) >= 3
+        if is_pagure_fork:
+            # fork/username/namespace/repo format
             username = split[1]
+            namespace_parts = split[2:-1]
+        else:
+            namespace_parts = split[:-1]
 
         return RepoUrl(
-            namespace=split[-2],
+            namespace="/".join(namespace_parts),
             repo=split[-1],
             username=username,
-            is_fork=is_fork,
+            is_fork=is_pagure_fork,
             hostname=parsed.hostname,
             scheme=parsed.scheme,
         )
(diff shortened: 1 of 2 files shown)


ogr/parsing.py
```diff
@@ -136,16 +136,21 @@ def parse_git_repo(potential_url: str) -> Optional[RepoUrl]:
         )
     if not username and len(split) >= 2:
         # path contains username/reponame
+        # or some/namespace/reponame
+        # or fork/username/some/namespace/reponame
 
-        is_fork = "fork" in split
-        if is_fork and len(split) == 4:
+        is_pagure_fork = "fork" in split
+        if is_pagure_fork and len(split) == 4:
```
mfocko, line 143: I wonder about https://gitlab.com/fork, since you can nest groups on gitlab...

ogr/parsing.py
```diff
@@ -136,16 +136,21 @@ def parse_git_repo(potential_url: str) -> Optional[RepoUrl]:
         )
     if not username and len(split) >= 2:
         # path contains username/reponame
+        # or some/namespace/reponame
+        # or fork/username/some/namespace/reponame
 
-        is_fork = "fork" in split
-        if is_fork and len(split) == 4:
+        is_pagure_fork = "fork" in split
```
csomh, line 142: Would checking for `split[0] == fork` be better?

ogr/parsing.py
csomh, line 143: Why '4'? What happens with https://pagure.io/fork/csomh/waiverdb ?

mfocko (approved):
I think I'll try to refactor it a bit on Friday.

ogr/parsing.py
lachmanfrantisek: Good point.

lachmanfrantisek: Fixed.

ogr/parsing.py
lachmanfrantisek: Yes, good point. Some historical relict from the Betka days... Fixed.

lachmanfrantisek: The current version checks only the first part to `fork` and https://gitlab.com/fork has not subgroups yet so it's probably safe.

ogr/parsing.py
csomh: How about adding a test for "https://host.name/namespace/fork"? A comment saying `a repo called "fork"` would be probably useful.

ogr/parsing.py
csomh: How about adding atest for `https://host.name/fork/username/repo`?

PR #467, opened 2020-09-17
Refactor `parse_git_repo`

mfocko:


ogr/parsing.py
```diff
@@ -125,6 +117,46 @@ def parse_git_repo(potential_url: str) -> Optional[RepoUrl]:
...
+    2) (same as above, but with ".git" in the end)
+    3) (same as the two above, but without "www.")
+    # all of the three above, but starting with "http://", "https://", "git://", "git+https://"
+    4) git@domain.com:foo/bar
+    5) (same as above, but with ".git" in the end)
+    6) (same as the two above but with "ssh://" in front or with "git+ssh" instead of "git")
+    7) pagure format of forks (e.g. domain.com/fork/username/namespace/project
+
+    Notably, the repo *must* have exactly username and reponame, nothing else and nothing
+    more. E.g. `github.com/<username>/<reponame>/<something>` is *not* recognized.
```
lachmanfrantisek, line 151: This is probably some old github-only note. Can you please update that so it matches the current state? (Change the wording to say that we don't support only main project URLs.)

lachmanfrantisek (commented):
Nice improvement!

I am only thinking if we can get rid of those tuples that are returned from the functions. Maybe filling the attributes of the final instance from the start?

mfocko:
> Nice improvement!
> 
> I am only thinking if we can get rid of those tuples that are returned from the functions. Maybe filling the attributes of the final instance from the start?

I was thinking if there is a reason it is not contained in the `RepoUrl`, probably make parsing `classmethod` and just call it from `parse_git_repo`? or something like `RepoUrl().parse(url)` and put those helper functions inside the class? I could get rid of the tuples that way

ogr/parsing.py
```diff
@@ -71,6 +71,98 @@ def __str__(self) -> str:
...
+        else:
+            namespace_parts = splits[:-1]
+
+        self.namespace = "/".join(namespace_parts)
+        self.repo = splits[-1]
+
+        return self
+
+    @classmethod
+    def _parse(cls, potential_url: str) -> Optional["RepoUrl"]:
```
lachmanfrantisek, line 152: It does not need to be private.

```suggestion
    def parse(cls, potential_url: str) -> Optional["RepoUrl"]:
```

lachmanfrantisek (commented):
I like that!

Just one small detail.

ogr/parsing.py
```diff
@@ -71,6 +71,98 @@ def __str__(self) -> str:
         repo_url_str += ")"
         return repo_url_str
 
+    def __repr__(self) -> str:
+        return str(self)
+
+    @staticmethod
+    def _prepare_url(potential_url: str) -> Optional[ParseResult]:
+        # Remove trailing '/' from 1-7 to ensure parsing works as expected.
+        if potential_url[-1] == "/":
```
csomh, line 80: I don't think this `if` is needed here.

ogr/parsing.py
```diff
@@ -71,6 +71,98 @@ def __str__(self) -> str:
         repo_url_str += ")"
         return repo_url_str
 
+    def __repr__(self) -> str:
+        return str(self)
+
+    @staticmethod
+    def _prepare_url(potential_url: str) -> Optional[ParseResult]:
+        # Remove trailing '/' from 1-7 to ensure parsing works as expected.
```
csomh, line 79: I know these comments were copied over from the old implementation, but I fail to understand them. What are the numbers about?

lachmanfrantisek: I think those numbers mean the examples in the function docstring. (But I agree with the update.)

ogr/parsing.py
```diff
@@ -71,93 +71,114 @@ def __str__(self) -> str:
...
+    2) (same as above, but with ".git" in the end)
+    3) (same as the two above, but without "www.")
+    # all of the three above, but starting with "http://", "https://", "git://", "git+https://"
+    4) git@domain.com:foo/bar
+    5) (same as above, but with ".git" in the end)
+    6) (same as the two above but with "ssh://" in front or with "git+ssh" instead of "git")
+    7) pagure format of forks (e.g. domain.com/fork/username/namespace/project
+
+    Repositories in nested groups (GitLab) or Pagure format of repos URLs
+    (e.g.`fork/username/namespace/repo`) are supported too.
```
lachmanfrantisek, line 179:
```suggestion
    7) pagure format of forks (e.g. domain.com/fork/username/namespace/project)
    8) nested groups on GitLab or Pagure (empty namespace is supported as well)
```

lachmanfrantisek (commented):
Just one docstring suggestion.Thanks!

lachmanfrantisek (approved):
Thanks!

You can maybe squash (or somehow get rid of) the last commits if we want to have nice commits in the history..;)

PR #477, opened 2020-10-08
Return `None` for user's email in Pagure

### LINKED ISSUE #126: get_email for GitUser

mfocko:
Chosen as a better resolution to unsupported API (previously raised an exception).

Closes #126

### DIFF
--- a/ogr/services/pagure/user.py
+++ b/ogr/services/pagure/user.py
@@ -22,6 +22,7 @@
 
 from typing import List
 
+from ogr.exceptions import OperationNotSupported
 from ogr.services import pagure as ogr_pagure
 from ogr.services.base import BaseGitUser
 from ogr.services.pagure.project import PagureProject
@@ -68,3 +69,9 @@ class PagureUser(BaseGitUser):
             )
             for fork in raw_forks
         ]
+
+    def get_email(self) -> str:
+        # Not supported by Pagure
+        raise OperationNotSupported(
+            "Pagure does not support retrieving of user's email address"
+        )


ogr/services/pagure/user.py
```diff
@@ -68,3 +68,7 @@ def get_forks(self) -> List["PagureProject"]:
             )
             for fork in raw_forks
         ]
+
+    def get_email(self) -> str:
+        # Not supported by Pagure
```
lachmanfrantisek, line 73: That's why I wanted to have exception. This information is lost.

We should have a consistent response for all not supported operations. (What is specific exception linking the issue if I am right.) I am not against change but we should be consistent.

What do you think? What is the way we want to go? `None` vs exception?

mfocko: I was thinking about making logger print out some message too.

Based on the discussion in the related issue, I'd say it's probably more non-disclosed than not implemented API? In the case of privacy concern (intentionally not implemented) it would make sense for me to return nothing. If it is really API problem, exception. Could be probably a good idea to check with the Pagure devs.

lachmanfrantisek: - We are speaking here about intentionally not-implemented methods -- the ones we can't implement because of the API.
- Similar problem can be with the arguments we can't process.
- We are mostly using `NonImplementedException` or `OperationNotSupported` (which might be a better fit):

https://github.com/packit/ogr/blob/e6c7ba7cab3332a6a969b7bd1e3c78cf8210609e/ogr/services/pagure/project.py#L314-L333
https://github.com/packit/ogr/blob/e6c7ba7cab3332a6a969b7bd1e3c78cf8210609e/ogr/services/pagure/project.py#L438-L441
https://github.com/packit/ogr/blob/e6c7ba7cab3332a6a969b7bd1e3c78cf8210609e/ogr/services/github/project.py#L194-L195
https://github.com/packit/ogr/blob/e6c7ba7cab3332a6a969b7bd1e3c78cf8210609e/ogr/services/github/project.py#L280-L289

- It can be anything but it should be same across the codebase.

> Could be probably a good idea to check with the Pagure devs.

I am not sure if we need that so much and since it can be treated as private info, I would not care about that-

csomh: After reading through this, I think I'll have to agree with @lachmanfrantisek's reasoning: when the forge doesn't support getting the email address, it would make more sense to raise an error (as it's an "exceptional" situation) than returning `None`. `OperationNotSupported` seems to be a good candidate, but I have no hard opinion on this.

I can think of 2 more reasons besides consistency, to do it this way:

1. It's explicit. A developer using this method might not realise a `None` return value, is also possible and would be more difficult to notice without an error.
2. `None` could also mean "no email for user", and so it would be difficult to tell this meaning apart from "forge does not support email".

mfocko:
@lachmanfrantisek @csomh Based on the discussion shall we close the PR with the related issue as `wontfix` (because of the API)?

csomh:
> @lachmanfrantisek @csomh Based on the discussion shall we close the PR with the related issue as `wontfix` (because of the API)?

Shouldn't we still have `PagureUser.get_email()` raise a `OperationNotSupported` or `NotImplementedError` with the explanation: "Pagure API does not support retrieving the email address of a user"?

csomh (approved):
I'm not sure if adding a test case for this would add some value :thinking: 

What do you think?

lachmanfrantisek (approved):
Thanks!

Can you please create an issue to change other occurrences to `OperationNotSupported` as well?

mfocko:
> I'm not sure if adding a test case for this would add some value 🤔
> What do you think?

probably not, we would just add a response file for `whoami` request and then check for exception

PR #479, opened 2020-10-14
Implement project.exists for GitLab

lachmanfrantisek:
Signed-off-by: Frantisek Lachman <flachman@redhat.com>

### DIFF
--- a/ogr/services/gitlab/project.py
+++ b/ogr/services/gitlab/project.py
@@ -24,7 +24,7 @@ import logging
 from typing import List, Optional, Dict, Set, Union
 
 import gitlab
-from gitlab.v4.objects import Project as GitlabObjectsProject
+from gitlab.v4.objects import GitlabGetError, Project as GitlabObjectsProject
 
 from ogr.abstract import (
     PullRequest,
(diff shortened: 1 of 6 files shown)


ogr/services/gitlab/project.py
```diff
@@ -120,6 +120,15 @@ def _construct_fork_project(self) -> Optional["GitlabProject"]:
             logger.debug(f"Project {self.repo}/{user_login} does not exist: {ex}")
         return None
 
+    def exists(self) -> bool:
+        try:
+            _ = self.gitlab_repo
+            return True
+        except GitlabGetError as ex:
+            if "404 Project Not Found" in str(ex):
```
mfocko, line 128: ~does it throw 404 for private repos too?~

tested it, it returns 401 and fails with exception as expected... what do you suggest?
- keep the exception
- return `False`, since it's not accessible
- return `True` as exists, but can't be accessed

mfocko (commented):
LGTM, i'll have a look at github later :)

ogr/services/gitlab/project.py
lachmanfrantisek: Good point!

I really can't decide...;)

It really returns a different error code for private repos? Is it secure?..;)

I can't decide, maybe `True` since it exists. WDYT?

mfocko: > Does it really return a different error code for private repos? Is it secure?..;)

it does 401 - unauthorized :D i expected 403

> I can't decide, maybe `True` since it exists. WDYT?

i think `True` is ok

lachmanfrantisek: Fixed.

mfocko (approved):
LGTM

PR #487, opened 2020-10-17
Add assignees argument while creating issue

### LINKED ISSUE #329: Assigning Issues to based on usernames.
Is it possible to assign Issues to particular user-names using the issue object by any chance?

#### Github
- github-api: https://developer.github.com/v3/issues/assignees/#add-assignees-to-an-issue
- pygithub: https://pygithub.readthedocs.io/en/latest/github_objects/Issue.html

#### Gitlab
- gitlab-api: https://docs.gitlab.com/ee/api/issues.html#edit-issue
- python-gitlab: https://python-gitlab.readthedocs.io/en/stable/gl_objects/issues.html
- example of the code: https://gitlab.fi.muni.cz/xlachma1/figitool/blob/master/figitool/figitool.py#L576
```python
            user_id = gitlab_instance.users.list(username=login)[0].id
            gitlab_mr.assignee_id = user_id
            gitlab_mr.save()
```

#### Pagure:
- https://src.fedoraproject.org/api/0/#issues
  - -> Create a new issue
  - -> Assign an issue

### TODO:

- [ ] add `assignee` argument to methods for creating the issue
  - in the `abstract.py`
  - for each implementation
  - add tests for it
- [ ] add the get/set property for the Issue classes
  - in the `abstract.py`
  - for each implementation
  - add tests for it

updated by @lachmanfrantisek

shreyaspapi:
fixes #329 

- [x] tests

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -189,7 +189,12 @@ class Issue:
 
     @staticmethod
     def create(
-        project: Any, title: str, body: str, private: Optional[bool] = None
+        project: Any,
+        title: str,
+        body: str,
+        private: Optional[bool] = None,
+        labels: Optional[List[str]] = None,
+        assignees: Optional[List[str]] = None,
     ) -> "Issue":
         """
         Open new Issue.
@@ -197,6 +202,8 @@ class Issue:
         :param project: Any
         :param title: str
         :param body: str
+        :param labels: list
+        :param assignees: list
         :return: Issue
         """
         raise NotImplementedError()
@@ -1024,6 +1031,7 @@ class GitProject:
         body: str,
         private: Optional[bool] = None,
         labels: Optional[List[str]] = None,
+        assignees: Optional[List[str]] = None,
     ) -> Issue:
         """
         Open new Issue.
@@ -1033,6 +1041,7 @@ class GitProject:
         :param body: str
         :param private: bool
         :param labels: [str]
+        :param assignees: [str]
         :return: Issue
         """
         raise NotImplementedError()
--- a/ogr/services/gitlab/issue.py
+++ b/ogr/services/gitlab/issue.py
@@ -66,6 +66,10 @@ class GitlabIssue(BaseIssue):
         return self._raw_issue.web_url
 
     @property
+    def assignees(self) -> list:
+        return self._raw_issue.assignees
+
+    @property
     def description(self) -> str:
         return self._raw_issue.description
 
@@ -96,10 +100,14 @@ class GitlabIssue(BaseIssue):
         body: str,
         private: Optional[bool] = None,
         labels: Optional[List[str]] = None,
+        assignees: Optional[List[str]] = None,
     ) -> "Issue":
         data = {"title": title, "description": body}
         if labels:
             data["labels"] = ",".join(labels)
+        if assignees:
+            data["assignee_ids"] = ",".join(assignees)
+
         issue = project.gitlab_repo.issues.create(data, confidential=private)
         return GitlabIssue(issue, project)
 
--- a/ogr/services/gitlab/project.py
+++ b/ogr/services/gitlab/project.py
@@ -425,9 +425,25 @@ class GitlabProject(BaseGitProject):
         body: str,
         private: Optional[bool] = None,
         labels: Optional[List[str]] = None,
+        assignees: Optional[List[str]] = None,
     ) -> Issue:
+
+        ids = []
+        for user in assignees or []:
+            users_list = self.service.gitlab_instance.users.list(username=user)
+
+            if not users_list:
+                raise GitlabAPIException(f"Unable to find '{user}' username")
+
+            ids.append(str(users_list[0].id))
+
         return GitlabIssue.create(
-            project=self, title=title, body=body, private=private, labels=labels
+            project=self,
+            title=title,
+            body=body,
+            private=private,
+            labels=labels,
+            assignees=ids,
         )
 
     def get_pr(self, pr_id: int) -> PullRequest:
--- a/tests/integration/github/test_issues.py
+++ b/tests/integration/github/test_issues.py
@@ -8,6 +8,9 @@ from ogr.exceptions import GithubAPIException
 
 @record_requests_for_all_methods()
 class Issues(GithubTests):
+    title = "This is a title"
+    description = "This is a description"
+
     def test_issue_list(self):
         issue_list = self.ogr_fork.get_issue_list()
         assert isinstance(issue_list, list)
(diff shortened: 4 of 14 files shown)


shreyaspapi:
Pagure does not support multiple assignees for an issue, should I add raise an exception if two or more items in the list?

lachmanfrantisek:
> Pagure does not support multiple assignees for an issue, should I add raise an exception if two or more items in the list?

Yes, that would be handy.

ogr/abstract.py
```diff
@@ -189,14 +189,21 @@ def __str__(self) -> str:
...
     @staticmethod
     def create(
-        project: Any, title: str, body: str, private: Optional[bool] = None
+        project: Any,
+        title: str,
+        body: str,
+        private: Optional[bool] = None,
+        labels: Optional[list] = None,
+        assignees: Optional[list] = None,
     ) -> "Issue":
```
lachmanfrantisek, line 198: Would be nice to add the type of the elements -- `Optional[List[str]]` probably.

Note that `list[str]` syntax is only for python3.9+

lachmanfrantisek (commented):
Nicely done, thanks!

As usual, some tests would be nice..;)

ogr/abstract.py
shreyaspapi: Hey, I tried it, the current `tox` config uses py36 and py37, Tests were failing there any way to silent that or has a way around to support both?

shreyaspapi: Might be some issues with my config🤷🏻‍♂️, I have fixed it!

ogr/services/gitlab/issue.py
```diff
@@ -96,10 +100,14 @@ def create(
         body: str,
         private: Optional[bool] = None,
         labels: Optional[List[str]] = None,
+        assignees: Optional[List[str]] = None,
     ) -> "Issue":
         data = {"title": title, "description": body}
         if labels:
             data["labels"] = ",".join(labels)
+        if assignees:
+            data["assignee_ids"] = ",".join(list(map(str, assignees)))
```
lachmanfrantisek, line 109: Is the `map` needed? From the type annotation, we should get list of strings or `None`.

ogr/services/gitlab/issue.py
```diff
@@ -112,6 +120,7 @@ def get(project: "ogr_gitlab.GitlabProject", id: int) -> "Issue":
 
     @staticmethod
     def get_list(
+        self,
```
lachmanfrantisek, line 123: Probably some rebase problem? Probably should not be in a static method.

ogr/services/gitlab/project.py
```diff
@@ -425,9 +425,20 @@ def create_issue(
         body: str,
         private: Optional[bool] = None,
         labels: Optional[List[str]] = None,
+        assignees: Optional[List[str]] = None,
     ) -> Issue:
+
+        ids = []
+        if assignees:
+            for user in assignees:
+                ids.append(self.service.gitlab_instance.users.list(username=user)[0].id)
```
lachmanfrantisek, line 434: Can we react nicely for the not-found user?

tests/integration/github/test_issues.py
```diff
@@ -39,6 +39,21 @@ def test_create_issue(self):
...
             project.create_issue(title=title, body=description, private=True)
 
+    def test_create_issue_with_assignee(self):
+        title = "This is an example issue"
+        description = "Issue description"
+        labels = ["label1", "label2"]
+        assignee = ["tunap"]
+        project = self.service.get_project(
+            namespace="BennettOSS", repo="Hacktoberfest19"
+        )
```
lachmanfrantisek, line 49: Would be nice if we can use some really testing project -- e.g. packit-service/hello-world. (Ideally also for the previous test...;) (And since we are in that, what about splitting `private=True` test to a separate test case?)

Thanks!!

lachmanfrantisek (commented):
Just a few suggestions.

Thanks!

ogr/services/gitlab/issue.py
shreyaspapi: Yup fixed this.

ogr/services/gitlab/issue.py
shreyaspapi: My bad forgot to convert to string in `def create_issue`

ogr/services/gitlab/project.py
```diff
@@ -425,9 +425,27 @@ def create_issue(
...
+        if assignees:
+            for user in assignees:
+                try:
+                    ids.append(
+                        str(
+                            self.service.gitlab_instance.users.list(username=user)[0].id
+                        )
+                    )
+                except gitlab.GitlabAPIException:
+                    raise GitlabAPIException(f"Unable to find '{user}' username")
```
lachmanfrantisek, line 441: Looks weird but I can't think of a better code for this..;)

Maybe this?

```suggestion
        ids = []
        for user in assignees or []:
            try:
                ids.append(
                    str(
                        self.service.gitlab_instance.users.list(username=user)[0].id
                    )
                )
            except gitlab.GitlabAPIException:
                raise GitlabAPIException(f"Unable to find '{user}' username")
```

lachmanfrantisek: Or what about this (looks like a really interesting problem..;):

```suggestion
        try:
            ids = [
                str(self.service.gitlab_instance.users.list(username=user)[0].id)
                for user in assignees or []
            ]
        except gitlab.GitlabAPIException:
            # ok, we should not use `user` here
            raise GitlabAPIException(f"Unable to find '{user}' username")
```

lachmanfrantisek: Last note..;)

Ca we check that the exception really means the not found user? (More specific type and/or content of the exception?)

lachmanfrantisek (commented):
Thanks!! I can't resist and have one more suggestion..;)

ogr/services/gitlab/project.py
shreyaspapi: The API does not throw any error if the user is not found.

shreyaspapi: But will throw out of index - `str(self.service.gitlab_instance.users.list(username=user)[0]`

shreyaspapi: How about this -

```
        ids = []
        for user in assignees or []:
            users_list = self.service.gitlab_instance.users.list(username=user)

            if users_list == []:
                raise GitlabAPIException(f"Unable to find '{user}' username")

            ids.append(users_list[0].id)
```

lachmanfrantisek (approved):
LGTM, thank you!

PR #488, opened 2020-10-20
Fix anonymous GitlabService

### LINKED ISSUE #481: GitLab service cannot be created without providing auth
ogr allows creating GitHub service instances without any authentication (provides read-only access)

Based on testing for #479 it's been found out that we cannot create GitLab instance without providing token (as we can with GitHub) *even though* it is supported by `python-gitlab` (https://python-gitlab.readthedocs.io/en/stable/api-usage.html#gitlab-gitlab-class)

mfocko:
- Allows no-auth read-only access to GitLab
  - authenticates only if provided token

Fixes #481

### DIFF
--- a/tests/integration/gitlab/test_service.py
+++ b/tests/integration/gitlab/test_service.py
@@ -4,6 +4,8 @@ from requre.online_replacing import record_requests_for_all_methods
 
 from tests.integration.gitlab.base import GitlabTests
 
+from ogr import GitlabService
+
 
 @record_requests_for_all_methods()
 class Service(GitlabTests):
@@ -80,3 +82,8 @@ class Service(GitlabTests):
             repo=name_of_the_repo, namespace=namespace_of_the_repo
         )
         assert project.gitlab_repo
+
+    def test_service_without_auth(self):
+        service = GitlabService()
+        assert service.gitlab_instance
+        assert service.get_project(namespace="gnachman", repo="iterm2").exists()
--- a/ogr/services/gitlab/service.py
+++ b/ogr/services/gitlab/service.py
@@ -50,7 +50,8 @@ class GitlabService(BaseGitService):
                 private_token=self.token,
                 ssl_verify=self.ssl_verify,
             )
-            self._gitlab_instance.auth()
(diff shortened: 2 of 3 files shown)


lachmanfrantisek (commented):
Nice!

What about some test so it's not so small change?..;)

mfocko:
> What about some test so it's not so small change?..;)

lemme have a look

tests/integration/gitlab/test_service.py
```diff
@@ -80,3 +82,7 @@ def test_project_create_in_the_group(self):
             repo=name_of_the_repo, namespace=namespace_of_the_repo
         )
         assert project.gitlab_repo
+
+    def test_service_without_auth(self):
+        service = GitlabService()
+        assert service.gitlab_instance
```
lachmanfrantisek, line 88: Can you make some real call so we know it really works? It does not touch anything now...

mfocko: Yes, I could... However I tested it with `master` and it fails with `401 Unauthorized` :D

mfocko:
recheck

tests/integration/gitlab/test_service.py
```diff
@@ -80,3 +82,8 @@ def test_project_create_in_the_group(self):
...
+        assert service.get_project(namespace="gnachman", repo="iterm2").exists()
```
lachmanfrantisek, line 89: Why this?..;)

lachmanfrantisek: Do we need to worry about the renaming of their repo? Or moving to `iterm2/iterm2` or something else?

lachmanfrantisek (approved):

tests/integration/gitlab/test_service.py
mfocko: they're already at version 3, so I wouldn't worry about it... I took it from the most stars and it was shorter than `inkscape/inkscape` :)

lachmanfrantisek: ..;)

PR #489, opened 2020-10-21
Adding delete project functionality

### LINKED ISSUE #215: Implement GitProject.delete()
Add method to project classes to delete a project.

AC:
- [ ] In each implementation, add the `delete` method to project class and implement it.
    - If possible/supported. (Document that if it is not possible and raise an exception with a good message.)
- [ ] Add tests for that.

Links:
- https://python-gitlab.readthedocs.io/en/stable/gl_objects/projects.html
- https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html
- https://pagure.io/api/0/ (not found the possible API call for that on the first look)

shreyaspapi:
fixes #215 

- [x] Added tests

### DIFF
--- a/tests/integration/github/test_generic_commands.py
+++ b/tests/integration/github/test_generic_commands.py
@@ -216,3 +216,9 @@ class GenericCommands(GithubTests):
 
     def test_is_not_private(self):
         assert not self.ogr_project.is_private()
+
+    def test_delete(self):
+        project = self.service.get_project(
+            repo="delete-project", namespace="shreyaspapi"
+        )
+        project.delete()
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -810,6 +810,10 @@ class GitProject:
         self.repo = repo
         self.namespace = namespace
 
+    def delete(self) -> None:
+        """Delete the project."""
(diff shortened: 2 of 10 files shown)


shreyaspapi:
I think I did something wrong here, will take a look.

tests/integration/github/test_generic_commands.py
```diff
@@ -216,3 +216,7 @@ def test_is_private(self):
 
     def test_is_not_private(self):
         assert not self.ogr_project.is_private()
+
+    def test_delete(self):
+        project = self.service.get_project(repo="temp-file", namespace="shreyaspapi")
```
lachmanfrantisek, line 221: Just a nit-pick -- what about using the same name as for others?

lachmanfrantisek (commented):
Just one small suggestion. Thanks a lot. It looks so easy and I am wondering why we don't have it already..;)

tests/integration/github/test_generic_commands.py
shreyaspapi: Done👍

lachmanfrantisek (approved):
:+1:

PR #495, opened 2020-11-12
Do not skip bug and security issues by stalebot

lachmanfrantisek:
Signed-off-by: Frantisek Lachman <flachman@redhat.com>


TomasTomecek (approved):
just to clarify for everyone: we want those issues to be bumped by stalebot so we're aware of them

PR #497, opened 2020-11-27
Add more tests to parsing and improve parsing

### LINKED ISSUE #496: ogr can't parse a subnamespace in gitlab's URL
```
In [1]: from ogr.parsing import parse_git_repo 

In [2]: parse_git_repo("git@gitlab.com:packit-service/src/libvirt.git")                        

In [3]: :( 
   ...:  
```

mfocko:
Will fix #496

### DIFF
--- a/ogr/parsing.py
+++ b/ogr/parsing.py
@@ -83,7 +83,7 @@ class RepoUrl:
         if "@" in potential_url:
             split = potential_url.split("@")
             if len(split) == 2:
-                potential_url = "http://" + split[1]
+                potential_url = "https://" + split[1]
             else:
                 # more @s ?
                 return None
@@ -92,7 +92,7 @@ class RepoUrl:
         if not potential_url.startswith(
             ("http://", "https://", "git://", "git+https://")
         ):
-            potential_url = "http://" + potential_url
+            potential_url = "https://" + potential_url
 
         return urlparse(potential_url)
 
@@ -100,14 +100,26 @@ class RepoUrl:
         self.hostname = parsed_url.hostname
         self.scheme = parsed_url.scheme
 
-    def _parse_username(self, parsed: ParseResult) -> None:
+    def _parse_username(self, parsed: ParseResult) -> bool:
         if ":" not in parsed.netloc:
-            return
+            return True
 
-        # e.g. domain.com:foo or domain.com:1234, where foo is username, but 1234 is port number
         split = parsed.netloc.split(":")
-        if split[1] and not split[1].isnumeric():
+        if len(split) > 2:
+            # in case there is port:namespace
+            return False
+
+        if not split[1]:
+            return True
+
+        if split[1] == "forks":
+            self.is_fork = True
+        elif not split[1].isnumeric():
+            # e.g. domain.com:foo or domain.com:1234,
+            # where foo is username, but 1234 is port number
             self.username = split[1]
+
+        return True
 
     @staticmethod
     def _prepare_path(parsed: ParseResult) -> Tuple[str, List[str]]:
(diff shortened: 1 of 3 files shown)


tests/unit/test_parsing.py
```diff
@@ -93,6 +93,68 @@
             ),
         ),
         ("https://fail@more@at@domain.com", None),
+        (
+            "git@gitlab.com:packit-service/src/libvirt.git",
+            RepoUrl(
+                repo="libvirt",
+                namespace="packit-service/src",
+                hostname="gitlab.com",
+                scheme="git+ssh"
```
mfocko, line 102: I'm aware of that, it should be `http`

mfocko:
@TomasTomecek @lachmanfrantisek I've added few more tests

tests/unit/test_parsing.py
```diff
@@ -93,6 +93,68 @@
...
         ),
         ("https://fail@more@at@domain.com", None),
+        (
+            "git@gitlab.com:packit-service/src/libvirt.git",
+            RepoUrl(
+                repo="libvirt",
+                namespace="packit-service/src",
+                hostname="gitlab.com",
+                scheme="git+ssh",
+            ),
```
TomasTomecek, line 103: I don't understand, how come this works here and fails for me locally? :D do I have old ogr?

TomasTomecek (commented):
Thanks for such a quick fix!

tests/unit/test_parsing.py
mfocko: it's not fixed yet :D there are 2 questions regarding it
1. do we want to keep user as in actual user? (right now we basically take the namespace, which is also a reason it failed for the url you put in the original issue)
2. shouldn't we default scheme to https (right now it's http)?

TomasTomecek: oh right, the tests are failing :facepalm: 

agreed, definitely default to https

my expectation would be, not sure if it makes sense:
```
assert namespace == "packit-service/src"
assert user == "packit-service"
```

mfocko: ok, I'll try to fix it and probably add comment that we mean user from the git-forge perspective (not the auth, since it is quite confusing that way)

Regarding the `http -> https`, are there any concerns for dependencies, e.g. some git-forge instance that is not running on https?

TomasTomecek: I'd say all run on https and probably even redirect to https

mfocko:
@TomasTomecek could you have a look? i'll rebase afterwards with changes if necessary

TomasTomecek (approved):
LGTM, very nice! thank you for taking care of this

ogr/parsing.py
```diff
@@ -92,22 +92,32 @@ def _prepare_url(potential_url: str) -> Optional[ParseResult]:
...
+            # in case there is port:namespace
+            return False
+
+        if split[1]:
+            if split[1] == "forks":
+                self.is_fork = True
+            elif not split[1].isnumeric():
+                # e.g. domain.com:foo or domain.com:1234,
+                # where foo is username, but 1234 is port number
+                self.username = split[1]
```
lachmanfrantisek, line 118:
```suggestion
        if not split[1]:
            return True

        if split[1] == "forks":
            self.is_fork = True
        elif not split[1].isnumeric():
            # e.g. domain.com:foo or domain.com:1234,
            # where foo is username, but 1234 is port number
            self.username = split[1]
```

Maybe?

lachmanfrantisek (approved):

ogr/parsing.py
lachmanfrantisek: Feel free to ignore..;)

mfocko: Fixed, still gotta do something about this on Friday

lachmanfrantisek (approved):
Still LGTM, thanks!

PR #498, opened 2020-11-27
Change exception for edited on GitLab's commit flag

### LINKED ISSUE #413: edited on Gitlab's commit flag
Follow-up of #344 (PR: #374)

API docs (https://docs.gitlab.com/ee/api/commits.html#post-the-build-status-to-a-commit):
> Adds or updates a build status of a commit.

Gitlab API suggests that setting commit flag can update existing one, which could mean this issue is blocked by API.

Updating could also mean:
1.  adding new commit flag that replaces old one
2.  adding new commit flag, but keeping the old one

Example output from docs doesn't contain commit flag with same name.

mfocko:
Fixes #413

### DIFF
--- a/ogr/services/gitlab/flag.py
+++ b/ogr/services/gitlab/flag.py
@@ -27,7 +27,7 @@ from typing import List, Union
 import gitlab
 
 from ogr.abstract import CommitFlag, CommitStatus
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
Thanks!

PR #499, opened 2020-11-27
Turn off requre in pre-commit CI

mfocko:
Turn off requre clean up in pre-commit CI since it times out.

not sure how much it changes running locally, I run manually and had to use `pre-commit run --hook-stage manual --all`

- [x] Check if is not skipped in zuul; gotta add `--hook-stage manual` to zuul too
- [x] What's wrong with the rebase??? (no network)
       There are no remotes in pre-commit CI?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 # HOWTO: https://pre-commit.com/#usage
 # pip3 install pre-commit
-# pre-commit install
+# pre-commit install -t pre-commit -t pre-push
 
 repos:
   - repo: https://github.com/psf/black
@@ -37,7 +37,9 @@ repos:
       - id: check-rebase
         args:
           - git://github.com/packit-service/ogr.git
+        stages: [manual, push]
   - repo: https://github.com/packit/requre
     rev: 0.2.2
     hooks:
       - id: requre-purge
+        stages: [manual, push]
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -37,7 +37,9 @@ repos:
       - id: check-rebase
         args:
           - git://github.com/packit-service/ogr.git
+        stages: [manual, push]
   - repo: https://github.com/packit/requre
     rev: 0.2.2
     hooks:
       - id: requre-purge
+        stages: [manual]
```
mfocko, line 45: @lachmanfrantisek `push` too at least? if we can't use `commit` because of the CI?

lachmanfrantisek: Yes, that would mean we run it on push and in zuul.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 # HOWTO: https://pre-commit.com/#usage
 # pip3 install pre-commit
-# pre-commit install
```
lachmanfrantisek, line 3: Can you please update config in other repositories as well, please?

mfocko:
- updated installation instructions (push hook is not installed automatically apparently)
- update makefile target to run it in manual hook stage
- rebase is run in zuul and before push
- requre is run in zuul and (should be) before push

lachmanfrantisek (approved):

.pre-commit-config.yaml
mfocko: yes

PR #514, opened 2020-12-15
PagureService: Make retries parametrizable

csomh:
This way users of a PagureService object can control how failures of API
calls are retried.

### DIFF
--- a/ogr/services/pagure/service.py
+++ b/ogr/services/pagure/service.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 
 import logging
-from typing import List, Optional
+from typing import List, Optional, Union
 
 import requests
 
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
Do we want to support this in all implementations?

(I am not sure about your plans, just want to provide as much unified interface as possible for all the forges.)

csomh:
> Thanks!
> Do we want to support this in all implementations?
> 
> (I am not sure about your plans, just want to provide as much unified interface as possible for all the forges.)

My immediate plan is to use this change in `dist2src` to retry some of the failures.

In general this would be nice, but as I see the `*Service` classes share very little in common, and I see no easy way to implement something similar for GitHub and GitLab. An issue like this would be nice for a student to research and implement, though.

lachmanfrantisek:
> *Service classes share very little in common

They should. Not the code, but the API.

PR #531, opened 2021-02-05
Add autogeneration of docs

mfocko:
# TODO

- [ ] decide between `pdoc` and `pdoc3` (they seem to be separate packages)
   oh, there's the tea https://github.com/mitmproxy/pdoc#pdoc-vs-pdoc3
- [ ] we currently use Sphinx markup that doesn't really work well with either of `pdoc`s
- [ ] since we use `__all__` it seems to be harder to navigate the docs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,36 @@
+name: Deploy documentation
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout branch
+        uses: actions/checkout@v2
+      - name: Build and push docs
+        run: |
+          git config user.name Packit
+          git config user.email user-cont-team@redhat.com
+
+          python3 -m pip install --upgrade pip
+          python3 -m pip install setuptools
+          python3 -m pip install .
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,30 @@
...
+        run: |
+          git config user.name Packit
+          git config user.email user-cont-team@redhat.com
+
+          pip3 install --user pdoc3 pygments
+
+          pdoc --html -o docs ogr
+          touch .nojekyll
+
+          git checkout -b gh-pages
```
csomh, line 26: Wait! Where is this going to be available? :)

TomasTomecek: Would it be better (or even feasible) to open a PR against the main branch with the docs and serve them from main?

mfocko: I was trying it out on my repo: https://mfocko.github.io/ogr/index.html

I'm also going to change the action trigger to _push to main_ and enable GitHub pages after the merge

lachmanfrantisek:
>     * we currently use Sphinx markup that doesn't really work well with either of `pdoc`s

This can be improved later.

> since we use __all__ it seems to be harder to navigate the docs

Same with this one.

lachmanfrantisek (approved):

PR #537, opened 2021-02-09
Implement retry mechanism for Github

mmuzila:
This update implements retry mechanism for Github requests.

Number of retry attempts can be set in `GithubService()` and  instances dictionary (used by `get_instances_from_dict()`), eg:
```
{                                                                   
    "github.com": {                                                 
                    "token": "abcd",                                            
                    "max_retries": "3", 
    }                                   
}   
``` 
This allows to enable request retries in packit/packit-service by adding  key `max_retries` to configuration (`authentication -> github -> max_retries`). I'm not sure If this is an acceptable solution, as far as it may be a bit confusing to set number of request retries not necessarily related to authentication in authentication section of config.

### DIFF
--- a/ogr/services/github/auth_providers/token.py
+++ b/ogr/services/github/auth_providers/token.py
@@ -1,7 +1,8 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
-from typing import Optional
+from typing import Optional, Union
+from urllib3.util import Retry
 
 import github
 
@@ -9,9 +10,9 @@ from ogr.services.github.auth_providers.abstract import GithubAuthentication
 
 
 class TokenAuthentication(GithubAuthentication):
-    def __init__(self, token: str, **_) -> None:
+    def __init__(self, token: str, max_retries: Union[int, Retry]=0, **_) -> None:
         self._token = token
-        self._pygithub_instance = github.Github(login_or_token=token)
+        self._pygithub_instance = github.Github(login_or_token=token, retry=max_retries)
 
     def __eq__(self, o: object) -> bool:
         return issubclass(o.__class__, TokenAuthentication) and (
@@ -32,5 +33,7 @@ class TokenAuthentication(GithubAuthentication):
         return self._token
 
     @staticmethod
-    def try_create(token: str = None, **_) -> Optional["TokenAuthentication"]:
-        return TokenAuthentication(token)
+    def try_create(
+        token: str = None, max_retries: Union[int, Retry]=0, **_
+    ) -> Optional["TokenAuthentication"]:
+        return TokenAuthentication(token, max_retries=max_retries)
--- a/ogr/services/github/auth_providers/token.py
+++ b/ogr/services/github/auth_providers/token.py
@@ -10,7 +10,7 @@ from ogr.services.github.auth_providers.abstract import GithubAuthentication
 
 
 class TokenAuthentication(GithubAuthentication):
-    def __init__(self, token: str, max_retries: Union[int, Retry]=0, **_) -> None:
+    def __init__(self, token: str, max_retries: Union[int, Retry] = 0, **_) -> None:
         self._token = token
         self._pygithub_instance = github.Github(login_or_token=token, retry=max_retries)
 
@@ -34,6 +34,6 @@ class TokenAuthentication(GithubAuthentication):
 
     @staticmethod
     def try_create(
-        token: str = None, max_retries: Union[int, Retry]=0, **_
+        token: str = None, max_retries: Union[int, Retry] = 0, **_
     ) -> Optional["TokenAuthentication"]:
         return TokenAuthentication(token, max_retries=max_retries)
--- a/ogr/services/github/service.py
+++ b/ogr/services/github/service.py
@@ -20,8 +20,9 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from typing import Optional, Type
+from typing import Optional, Type, Union
 
+from urllib3.util import Retry
 import github
 import github.GithubObject
 from github import (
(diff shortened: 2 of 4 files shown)


ogr/services/github/auth_providers/token.py
```diff
@@ -9,9 +9,9 @@
 
 
 class TokenAuthentication(GithubAuthentication):
-    def __init__(self, token: str, **_) -> None:
+    def __init__(self, token: str, max_retries=0, **_) -> None:
         self._token = token
-        self._pygithub_instance = github.Github(login_or_token=token)
+        self._pygithub_instance = github.Github(login_or_token=token, retry=max_retries)
```
lachmanfrantisek, line 14: What about being able to set the `Retry` object as well as it is possible in case of Pagure:

https://github.com/packit/ogr/blob/135dad28cc85f2f91b8e3b133e3db2c2210d3834/ogr/services/pagure/service.py#L50

lachmanfrantisek: Actually, both here and in the service class.

ogr/services/github/service.py
```diff
@@ -59,6 +60,7 @@ def __init__(
         github_app_private_key_path: str = None,
         tokman_instance_url: str = None,
         github_authentication: GithubAuthentication = None,
+        max_retries=0,
```
lachmanfrantisek, line 63: What about having some constant for this and use here and in the `PagureService` as well?

lachmanfrantisek (commented):
I'm really surprised! Nice and easy solution.

ogr/services/github/auth_providers/token.py
mmuzila: You're right, it could be handy. Though it must be used wisely.

ogr/services/github/service.py
mmuzila: The question is, if we want to have the same retry policy for `GithubService` and `PagureService`.

It shouldn't be a problem to have a constant defined for both services, but in that case I think, it should  be an integer rather than  `urllib3.util.Retry`. Currently the `Retry` object for `GithubService` is constructed [here](https://github.com/packit/ogr/blob/39a6b4985490d719e2a0f18d69b0034e5a949fb5/ogr/services/github/service.py#L76), and quite defensive strategy was chosen for now. It retries only in specific cases (only HTTP status 401, 403, 500; only certain HTTP methods; only certain error types).

lachmanfrantisek: OK, it makes sense to have different defaults.

ogr/services/github/service.py
```diff
@@ -71,13 +73,25 @@ def __init__(
...
         self.authentication = github_authentication
 
+        if isinstance(max_retries, Retry):
+            self._max_retries = max_retries
+        else:
+            self._max_retries = Retry(
+                total=int(max_retries),
+                read=0,
+                method_whitelist=["DELETE", "GET", "PATCH", "POST", "PUT"],
+                status_forcelist=[500, 403, 401],
```
TomasTomecek, line 83: 403, 500 - definitely agree, ideally all variants of 500 (especially 503)

for 401 - what exactly is the use case, are we actually getting 401 somewhere which succeeds moments later?

TomasTomecek (commented):
very nice implementation!

ogr/services/github/service.py
```diff
@@ -71,13 +73,25 @@ def __init__(
...
+                raise_on_status=False,
+            )
+
```
csomh, line 86: I don't think this is required. `github.Github` also takes `int` or `Retry` as an argument, and `Retry` has some [sane defaults](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry). Or do you want something different then that?

mmuzila: I read mentioned `Retry` (documentation)[https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.Retry] but I wasn't satisfied with its defaults. I wanted to specify: 
- raise_on_status=False - otherwise `urllib3.exceptions.MaxRetryError` would be raised when max number of retries was reached. I wanted the retry mechanism to have minimal effects on the rest of code.
-  read=0 in order not to retry a call that could have side-effects
- status_forcelist - retry only on specific status codes

I was a quite conservative when specifying conditions when to retry - I didn't want to break anything. Other cases can be easily added in future if needed.

ogr/services/github/service.py
mmuzila: > for 401 - what exactly is the use case, are we actually getting 401 somewhere which succeeds moments later?

It was [requested](https://github.com/packit/packit-service/issues/830) to retry on `BadCredentialsException`. It is [raised when HTTP status is 401](https://github.com/PyGithub/PyGithub/blob/6900f187b53711d7e9e43fcadede2e44d7d12f7d/github/Requester.py#L412). AFAIK it shouldn't break [HTTP authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication) (but of course the retry mechanism will slow it down).

ogr/services/github/service.py
csomh: I see, thank you for the explanation! Could you maybe add a comment explaining why `method_whitelist` and `status_forcelist` is different thatn the defaults. The rest of the values are kinda self-explanatory, but these two are not that ovious.

ogr/services/github/service.py
TomasTomecek: understood and makes sense, let's also keep 401

TomasTomecek (approved):
LGTM, very nice work!

ogr/services/github/service.py
mmuzila: @csomh, do you agree with changes (added comments, and 503 status code) done in 70b992e ?

csomh (approved):
:+1:

jpopelka (approved):
It's not WIP anymore, is it?

PR #542, opened 2021-02-18
Fix getting latest release

### LINKED ISSUE #540: `get_latest_release` when no release exists
```py
    def get_latest_release(self) -> Release:
```

In case that no release exists, I'd expect to get `None` instead of being smacked in face with exception.

- GitHub throws 404 Exception
- GitLab throws IndexError

Action items:
- [ ] Switch return type from `Release` to `Optional[Release]`
- [ ] Update implementation to check for such situation

Could be used in refactor for packit/packit-service#985, currently is a blocker, since we need to check non-existing release there.

mfocko:
In case there is no existing release, instead of failing with an exception
return `None`.

Fixes #540

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -1173,7 +1173,7 @@ class GitProject(OgrAbstractClass):
         """
         raise NotImplementedError()
 
-    def get_latest_release(self) -> Release:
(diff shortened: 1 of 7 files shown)


lachmanfrantisek (approved):
Thanks!

PR #549, opened 2021-03-04
Allow ignoring custom instances when creating a project

csomh:
The current behaviour of 'ogr.get_project()', when a list of
'custom_instances' (already created GitService objects) is provided, is
to raise an OgrException if none of these matches the project URL.

Introduce a 'force_custom_instance' flag, defaulting to True to keep
backwards compatibility, in order to support the use case where the user
would like to retrieve an existing service *or create a new one* if none
is matching in the 'custom_instances' list.

### DIFF
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -69,6 +69,7 @@ def get_project(
     url,
     service_mapping_update: Dict[str, Type[GitService]] = None,
     custom_instances: Iterable[GitService] = None,
+    force_custom_instance: bool = True,
     **kwargs,
 ) -> GitProject:
     """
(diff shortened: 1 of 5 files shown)


csomh:
Added one more commit to fix some issues when testing in a container.

ogr/factory.py
```diff
@@ -97,11 +101,12 @@ def get_project(
                 service = service_inst
                 break
         else:
-            raise OgrException(
-                f"Instance of type {kls.__name__} "
-                f"matching instance url '{url}' was not provided."
-            )
-    else:
+            if force_custom_instance:
```
mfocko, line 104: could be merged into `elif` :)

mfocko (approved):
LGTM :)

ogr/factory.py
lachmanfrantisek: That's `for-else` and `for-elif` is not possible..;)

lachmanfrantisek (approved):

ogr/factory.py
mfocko: oh my, there's `for` above it :D it didn't seem like that in diff :D

PR #552, opened 2021-03-08
pagure: enable creating PRs from fork via fork_username

### LINKED ISSUE #551: authentication: support alternative server for git and git-forge API (src.fp.o)
This is gonna be a fun one: I tried to use upsint to create a PR in src.fedoraproject.org/rpms but failed because there are two different servers for the forge API and a git server:

I use SSH:
```
ssh://ttomecek@pkgs.fedoraproject.org/rpms/python-ogr.git
```

But pagure API is at
```
https://src.fedoraproject.org/rpms/python-ogr.git
```

I'm wondering if it's possible to set up an authentication entry which would recognize pkgs.fp.o but route API requests to src.fp.o. Tried these but none of them worked with SSH:
```
  fedora-dist-git:
    type: pagure
    instance_url: "https://src.fedoraproject.org/"
    token: ...
  fedora-dist-git2:
    type: pagure
    instance_url: "https://pkgs.fedoraproject.org/"
    token: ...
```

Worst case I could hard-code the logic to upsint though I think it would be useful to have in ogr itself.

TomasTomecek:
PagureProject.create_pr dumped fork_username which made PR creation from
a fork by passing this kwarg not working - one had to create the PR from
an actual fork instance instead of making it the parent repo.

This commit:
* makes sure the fork_username is passed to PagurePullRequest.create
* and also sets correct `repo_from[*]` keys for the API request to
  create a PR

Worked for: https://src.fedoraproject.org/rpms/packit/pull-request/125
And also this PR -_-

Fixes #551

~~Please let me know what tests I should write.~~

* [x] Wrote a test case for this.

### DIFF
--- a/ogr/services/pagure/project.py
+++ b/ogr/services/pagure/project.py
@@ -272,6 +272,7 @@ class PagureProject(BaseGitProject):
             body=body,
             target_branch=target_branch,
             source_branch=source_branch,
+            fork_username=fork_username,
         )
 
     @if_readonly(return_function=GitProjectReadOnly.fork_create)
--- a/ogr/services/pagure/pull_request.py
+++ b/ogr/services/pagure/pull_request.py
@@ -175,6 +175,16 @@ class PagurePullRequest(BasePullRequest):
 
             # running the call from the parent project
             caller = caller.parent
+        elif fork_username:
+            fork_project = project.service.get_project(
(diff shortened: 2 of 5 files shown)


TomasTomecek:
-_-

jpopelka (approved):

PR #553, opened 2021-03-09
Workflow for uploading to PyPI when a release is created

jpopelka:


.github/workflows/pypi-publish.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: |
+          python -m pip install build
+          python -m build --sdist --wheel
+
+      - name: Publish 📦 to PyPI
+        # https://github.com/pypa/gh-action-pypi-publish
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_API_TOKEN }}
+          verbose: true
```
TomasTomecek, line 29: will only we be able to see the output from this step?

also, whose API token is set here?

TomasTomecek (commented):
you were right, short and easy

it seems that release-bot's time is over if it's this easy to set up releasing to PyPI

.github/workflows/pypi-publish.yml
jpopelka: > will only we be able to see the output from this step?

From what I checked (for [example](https://github.com/systemd/systemd/actions)) everyone can see the logs.

It [adds](https://github.com/pypa/gh-action-pypi-publish/blob/master/twine-upload.sh#L44) `--verbose` to `twine upload` and I don't think `twine` is so stupid to output the password even with `--verbose`.
But if it is, I'll simply revoke the password and remove this flag.

> whose API token is set here?

Good question. [usercont-release-bot](https://pypi.org/user/usercont-release-bot) and I've been thinking about whether to use my account or create a completely new "packit-bot" account.
Also, I know I should comment it somewhere, but not sure where, I wish there was some possibility to add a comment to [secrets](https://github.com/packit/ogr/settings/secrets/actions).

jpopelka: Added the comment. Can someone approve?

csomh (approved):

PR #554, opened 2021-03-10
Install rpmautospec-rpm-macros on Fedora only

jpopelka:


TomasTomecek:
pity they don't build it for epel, would be nice to ask them about it

TomasTomecek (approved):

Issue #559, opened 2021-03-15
mypy cannot use the type annotation


Roming22:
The package is not PEP-561 compliant (see https://mypy.readthedocs.io/en/latest/installed_packages.html#creating-pep-561-compatible-packages).

The file `py.typed` need to be added at the root of the ogr package.

lachmanfrantisek:
Good to know, thanks for creating the issue and sending us the PR.

mfocko:
closed by 33f4be5, right?

jpopelka:
yes, #560

PR #560, opened 2021-03-15
Export types in the package (PEP-561)

Roming22:
This is the fix for [issue #559](https://github.com/packit/ogr/issues/559)

The file py.typed need to be added at the root of the ogr package for
the package to be PEP-561 compliant and have mypy be able to use the
type information when importing the package in other projects.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,6 +50,9 @@ setup_requires =
 testing =
     pytest
 
+[options.package_data]
+* = py.typed
+
 [options.packages.find]
 exclude =
     tests*


jpopelka:
Per [1](https://www.python.org/dev/peps/pep-0561/), [2](https://stackoverflow.com/questions/62979389/is-there-a-best-practice-to-make-a-package-pep-561-compliant) the `py.typed` should be included in [setup.cfg](https://github.com/packit/ogr/blob/main/setup.cfg)

jpopelka (approved):
LGTM, however, I'm not sure how to verify that it works as expected.
I built & installed a wheel and checked the `site-packages/ogr.*.dist-info/`, but don't see the `py.typed` mentioned anywhere.

But we can merge and see. @Roming22 can we merge now?

PR #561, opened 2021-03-16
Authentication type: `gitlab` results in `TypeError`

mfocko:


TomasTomecek (approved):
This will totally solve my problem, thank you!

I'm personally not fond of this "argument eating" so if you think this is a good idea, I'm not gonna block the merge :)

I'd probably at least logger.warn about arguments which were not processed.

mfocko:
> This will totally solve my problem, thank you!
> 
> I'm personally not fond of this "argument eating" so if you think this is a good idea, I'm not gonna block the merge :)
> 
> I'd probably at least logger.warn about arguments which were not processed.

It's done in each of the services; I still need to rebase, so if you insist I can change it before pushing :)

TomasTomecek:
> It's done in each of the services; I still need to rebase, so if you insist I can change it before pushing :)

I approved, you're welcome to merge.

TomasTomecek:
someone is cautious :)

mfocko:
> someone is cautious :)

Better than debug :D

PR #571, opened 2021-04-21
Unify behaviour for get_file_content for GitHub

mfocko:
Catch `GithubException` in GithubProject.get_file_content as we do
for GitLab and Pagure to:
a) unify the behaviour
b) ignore unwanted 404 related to:
       https://sentry.io/organizations/red-hat-0p/issues/1809076278


lachmanfrantisek (approved):
Thanks for improving that!

PR #572, opened 2021-04-23
0.24.0-release

mmuzila:


jpopelka (commented):
We agreed on not using markdown in the CHANGELOG.md, because of [copy_upstream_release_description: true](https://github.com/packit/ogr/blob/main/.packit.yaml#L11), which [copies the changelog entry into downstream spec %changelog](https://packit.dev/docs/configuration/#copy_upstream_release_description), where we don't want to have markdown.

jpopelka (approved):

TomasTomecek (approved):

PR #574, opened 2021-04-26
Use GitlabGetError from gitlab.exceptions

### LINKED ISSUE #573: ImportError: cannot import name 'GitlabGetError' from 'gitlab.v4.objects'
After the latest python-gitlab release, this error showed up in [packit tests](https://github.com/packit/packit/pull/1213#issuecomment-826668194).

csomh:
Imports from other 'gitlab' modules might work, if they in turn import
the exception, but things will stop working as soon as this stops being
true :D

Fixes #573.


csomh:
recheck

mfocko:
did they actually agree on changing the API in both gitlab and github? 🤔

csomh:
> did they actually agree on changing the API in both gitlab and github?  (github mail is failing because it isn't dictionary anymore but `EmailData` object ...)

I might need to regenerate the test data... will check it tomorrow.

TomasTomecek:
nice, such an easy fix and only two requre test failures - should be doable

I wonder if it's worth to do a new patch release (or maybe only a PyPI release so that our tests in packit would pass)

mfocko:
> nice, such an easy fix and only two requre test failures - should be doable
> 
> I wonder if it's worth to do a new patch release (or maybe only a PyPI release so that our tests in packit would pass)

except the github email, that looks like change in PyGithub; it probably needs to be backward compatible because of RPM

lachmanfrantisek:
@csomh Thanks for taking a look. The amount of failed tests across our projects is really huge for such a small change..;-)

(I hope they haven't put anything new to the evening's [fix release](https://github.com/python-gitlab/python-gitlab/releases/tag/v2.7.1)...)

lachmanfrantisek (approved):
Thanks!

TomasTomecek (approved):

PR #581, opened 2021-05-10
ignored kwargs: set log to debug and only log if set

TomasTomecek:
I'm tired of

    Ignored keyword arguments: {}

and since this change was requested by me, finally decided to improve
the behavior:

1. log only when some kwargs are set

2. set log to debug

### DIFF
--- a/ogr/services/github/service.py
+++ b/ogr/services/github/service.py
@@ -99,7 +99,8 @@ class GithubService(BaseGitService):
                 max_retries=self._max_retries,
             )
 
-        logger.warning(f"Ignored keyword arguments: {kwargs}")
+        if kwargs:
+            logger.warning(f"Ignored keyword arguments: {kwargs}")
 
     def __set_authentication(self, **kwargs):
         auth_methods = [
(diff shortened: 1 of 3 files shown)


ogr/services/github/service.py
```diff
@@ -99,7 +99,8 @@ def __init__(
                 max_retries=self._max_retries,
             )
 
-        logger.warning(f"Ignored keyword arguments: {kwargs}")
+        if kwargs:
+            logger.debug(f"Ignored keyword arguments: {kwargs}")
```
lachmanfrantisek, line 103: I am thinking if it is really not worth a `warning`? Since it's printed only for non-empty `kwargs`, I would go with `warning`.

TomasTomecek: I definitely saw examples where kwargs contained actual data: I recall username and max_retries.

lachmanfrantisek: And that should be fixed on the called side I think. @mfocko WDYT?

TomasTomecek: I'll change it back to warning: you're right we'll be able to correct the usage then

lachmanfrantisek (approved):

TomasTomecek:
recheck

csomh:
Thank you for this!

PR #583, opened 2021-05-13
Refactoring

mfocko:


mfocko:
passing locally :hmm:

ogr/services/github/project.py
```diff
@@ -277,20 +277,20 @@ def _get_collaborators_with_permission(self) -> dict:
...
         self,
         status: IssueStatus = IssueStatus.open,
         author: Optional[str] = None,
         assignee: Optional[str] = None,
         labels: Optional[List[str]] = None,
     ) -> List[Issue]:
-        return GithubIssue.get_list(
-            project=self, status=status, author=author, assignee=assignee, labels=labels
-        )
+        pass
```
lachmanfrantisek, line 288: Is mypy `ok` with that? Will it still check if the attributes do not match?

ogr/services/github/pull_request.py
```diff
@@ -167,8 +167,8 @@ def __get_fork(fork_username: str, repo: _GithubRepository) -> _GithubRepository
         return forks[0]
 
     @staticmethod
-    def get(project: "ogr_github.GithubProject", id: int) -> "PullRequest":
-        pr = project.github_repo.get_pull(number=id)
+    def get(project: "ogr_github.GithubProject", pr_id: int) -> "PullRequest":
+        pr = project.github_repo.get_pull(number=pr_id)
```
lachmanfrantisek, line 171: Hmm. Change of the API? Can we safely rename that?

ogr/services/gitlab/issue.py
```diff
@@ -102,21 +102,30 @@ def create(
...
+                raise GitlabAPIException(f"Unable to find '{user}' username")
+
+            assignee_ids.append(str(users_list[0].id))
+
         data = {"title": title, "description": body}
         if labels:
             data["labels"] = ",".join(labels)
         if assignees:
-            data["assignee_ids"] = ",".join(assignees)
+            data["assignee_ids"] = ",".join(assignee_ids)
```
lachmanfrantisek, line 118: Can we have it in a separate commit, please?

lachmanfrantisek (approved):
What about merging this independently to check runs?

ogr/services/github/project.py
mfocko: Seems to be catching it, just tested it

ogr/services/github/pull_request.py
mfocko: It's internal, I doubt it is used independently of the project. It would be PITA to use, since you need to provide project separately.

ogr/services/gitlab/issue.py
mfocko: Sure, no problem.

lachmanfrantisek: Sorry that I haven't made it clear. I meant only the fix for assignees. (So it can be found in the history that we've changed something.)

lachmanfrantisek: Aha, ok this is just a refactor. Not a fix.

mfocko: Yeah, I just moved the code from the project to class where it should be :D But I think it's nice to have it separate, since it's bigger change than just removing code and decorating

mfocko:
recheck

Issue #586, opened 2021-05-20
Adding assignees for issue that already exists


KPostOffice:
Hi all!

Really enjoying working with `ogr`. Would it be possible to add the functionality to add assignees to an issue which already exists?

Refs:
* https://docs.github.com/en/rest/reference/issues#add-assignees-to-an-issue
* https://pygithub.readthedocs.io/en/latest/github_objects/Issue.html#github.Issue.Issue.add_to_assignees
* attached issue shows how to add assignee for GitLab
* https://src.fedoraproject.org/api/0/#issues-tab

See also: https://github.com/packit/ogr/pull/487

lachmanfrantisek:
Hi @KPostOffice, glad you find `ogr` useful!

That's a valid feature request and not hard to implement.

Since you've already linked the documentation and gone through the code, will you be interested in contributing this? (We can help you with that.) Otherwise, I can't tell you an approximation of when this will be ready.

Regarding the implementation, we need to:
* [ ] Add the abstract `add_assignee` method to the `Issue` class:
    * https://packit.github.io/ogr/abstract.html#ogr.abstract.Issue
    * https://github.com/packit/ogr/blob/main/ogr/abstract.py#L139
* [ ] Implement the method for `Github`/`GitLab`/`Pagure`.
* [ ] Add tests for this (https://github.com/packit/ogr/blob/main/CONTRIBUTING.md#testing).

KPostOffice:
Yeah, I can contribute this. It seems pretty straight forward. :+1:

PR #589, opened 2021-05-25
add function for adding assignees to issues

KPostOffice:
Closes: https://github.com/packit/ogr/issues/586

### DIFF
--- a/ogr/services/gitlab/issue.py
+++ b/ogr/services/gitlab/issue.py
@@ -67,7 +67,10 @@ class GitlabIssue(BaseIssue):
 
     @property
     def assignees(self) -> list:
-        return self._raw_issue.assignees
+        try:
+            return self._raw_issue.assignees
+        except AttributeError:
+            return None  # if issue has no assignees, the attribute is not present
 
     @property
     def description(self) -> str:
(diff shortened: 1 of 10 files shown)


lachmanfrantisek (commented):
Thanks for the PR, the code looks really good!

Regarding the tests, the code looks good but we need to generate the yaml files with the responses.
* As you might know from the [contribution guide](https://github.com/packit/ogr/blob/main/CONTRIBUTING.md#testing), If you run the tests locally, you need to provide github/gitlab/pagure tokens via env.var. and run the tests. This will touch the real services and save the responses to the yaml files. Those files need to be committed as well so we don't need to use secrets and touch the real services in CI. (Basically, it records the communication and removes the sensitive values.)
* Since it touches the real services, your account need to have the permissions to do assigning -- e.g. on GitHub, I can give you `triage` permission if you want.
* I can also generate the response files for you. (I understand it's tricky for the first time.)

Thanks!

KPostOffice:
@lachmanfrantisek Do I just use my own personal access tokens for github/gitlab/pagure?

mfocko:
> Do I just use my own personal access tokens for github/gitlab/pagure?

@KPostOffice yes, but I think you need write access to the repository to set assignees... I think there is some test I have created before that used user-agnostic repository... lemme have a look... `test_add_user` in `generic_commands`... I think you could use a skeleton like that

KPostOffice:
@lachmanfrantisek PTAL

ogr/services/gitlab/issue.py
```diff
@@ -171,3 +174,18 @@ def add_label(self, *labels: str) -> None:
...
+            users = self.project.service.gitlab_instance.users.list(  # type: ignore
+                username=assignee
+            )
+            if not users:
+                raise GitlabAPIException(f"Unable to find '{assignee}' username")
+            uid = str(users[0].id)
+            if uid not in a:
+                a.append(str(users[0].id))
+
+        self._raw_issue.assignee_ids = a
```
lachmanfrantisek, line 190: Can we have a better name for the assignee list, please?

lachmanfrantisek (commented):
Thanks!
Just one small suggestion, otherwise LGTM.

Can you please do `rebase` instead of `merge` so we have only one new commit that sits on top of the current `main` branch? (Let me know if you need some help with that.)

Thank you for working on this!

KPostOffice:
@lachmanfrantisek I addressed both comments. It should push on top of master nicely now.

lachmanfrantisek (approved):
Nice, thanks for the contribution!

PR #593, opened 2021-05-27
Make 'make check' more consistent across projects

bcrocker15:
We have decided to remove the prepare-check and prepare-build
targets entirely, relying instead on strongly-worded recommendations
in CONTRIBUTING.md and README.md.

Fixes ogr issue #579.

Signed-off-by: Ben Crocker <bcrocker@redhat.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,36 +32,23 @@ Here are some links to the documentation that could be helpful when contributing
 
 Tests are stored in the [tests](/tests) directory.
 
-Running tests locally:
+We recommend running tests in a container. You need to have `podman` or
+`docker` installed for this.
 
-```
-make prepare-check && make check
-```
+To build the test image run:
 
-The following packages are installed by 'make prepare-check'
-if they are not installed already:
+    make build-test-image
 
-    python3-flexmock
-    python3-requre
+You can run the test now with:
 
-Or you can run tests in a container:
-
-    make prepare-build
-    make build
     make check-in-container
 
-The following packages are installed by 'make prepare-build'
-if they are not installed already:
-
-    podman
-    ansible-bender
-
-The 'make build' command builds a local image with all the dependencies using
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -2,26 +2,10 @@ BASE_IMAGE := fedora:latest
...
-else
-    PACKAGE_CHECKER := :
-    PACKAGE_INSTALLER := :
-endif
-
-prepare-build:
-	if ! $(PACKAGE_CHECKER) ansible-bender ; then sudo $(PACKAGE_INSTALLER) ansible-bender ; fi
-	if ! $(PACKAGE_CHECKER) podman ; then sudo $(PACKAGE_INSTALLER) podman ; fi
 
 build: recipe.yaml
```
csomh, line 6: I think this should be renamed to `build-test-image` in order to be consistent with the rest of the projects. Doing so might also reduce some of the confusion around what the built image is going to be used for (In other projects, `make build` would built a "service" image, which would be used in a deployment. OGR being a library, doesn't have something like this, obviously.)

CONTRIBUTING.md
```diff
@@ -34,28 +34,26 @@ Tests are stored in the [tests](/tests) directory.
...
+installed beforehand:
 
     podman
     ansible-bender
 
+Then:
+
+    make build
+    make check-in-container
+
```
csomh, line 56: Please adjust the wording of this part to resemble [the one here](https://github.com/packit/packit-service-fedmsg/pull/43/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R9-R30).

csomh:
Changes in `files/zuul-tests.yaml` and `recipe.yaml` are not needed.

csomh (approved):
I think this is not a draft anymore :smiley: 

Good to be merged! Thanks!

PR #613, opened 2021-07-13
Add GithubPullRequest.patch property

nikromen:
Implementated for GithubPullRequest.

It returns bytes in case there's some binary blob in the patch.
It uses requests to get content from patch's URL because I couldn't find in code any function which will get content from a given URL.

### DIFF
--- a/ogr/services/github/pull_request.py
+++ b/ogr/services/github/pull_request.py
@@ -22,6 +22,7 @@
 
 import datetime
 import logging
+import requests
 from typing import Optional, List, Union
 
 from github import UnknownObjectException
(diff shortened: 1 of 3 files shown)


mfocko (commented):
LGTM :)

jpopelka (approved):
Thanks!
[Squash/fixup](https://www.mikulskibartosz.name/git-fixup-explained/) the commits, please.

ogr/services/github/pull_request.py
```diff
@@ -101,6 +102,17 @@ def labels(self) -> List[GithubLabel]:
...
+    @property
+    def patch(self) -> bytes:
+        response = requests.get(self._raw_pr.patch_url)
+
+        if not response.ok:
+            raise GithubAPIException(
+                f"Couldn't get patch from {self._raw_pr.patch_url} because {response.reason}."
+            )
+
+        return response.content
```
TomasTomecek, line 114: this is definitely a correct thing to do since we don't know what's inside and might not be able to decode the content

TomasTomecek (approved):

jpopelka:
Ok, looks like we're done here.
@nikromen once you're ready for merge, just add the `mergeit` label - click the wheel next to `Labels` on top(-ish) right of the page and add it there.

PR #618, opened 2021-07-20
Avoid exc when conclusion of Github check run is None

lbarcziova:


ogr/services/github/check_run.py
```diff
@@ -156,7 +156,11 @@ def started_at(self, started_at: datetime.datetime) -> None:
     @property
     def conclusion(self) -> GithubCheckRunResult:
```
mfocko, line 157:
```suggestion
    def conclusion(self) -> Optional[GithubCheckRunResult]:
```

mfocko (approved):

PR #621, opened 2021-08-02
Gitlab: create forks in "<username>/forks" namespace (if exists)

jpopelka:
Gitlab has (sub)groups so we don't have to pollute the authenticated user's main projects page, but can create the forks in a separate forks group - if exists (like [this one](https://gitlab.com/packit-service/forks)).
If there's no such group, it works as previously, i.e. creates/uses forks in the user's main group/namespace.

### DIFF
--- a/ogr/services/gitlab/project.py
+++ b/ogr/services/gitlab/project.py
@@ -93,17 +93,23 @@ class GitlabProject(BaseGitProject):
             and self.service == o.service
         )
 
-    def _construct_fork_project(self) -> Optional["GitlabProject"]:
-        user_login = self.service.user.get_username()
+    def _get_project_in_namespace(self, namespace: str) -> Optional["GitlabProject"]:
         try:
             project = GitlabProject(
-                repo=self.repo, service=self.service, namespace=user_login
+                repo=self.repo, service=self.service, namespace=namespace
             )
             if project.gitlab_repo:
                 return project
         except Exception as ex:
-            logger.debug(f"Project {self.repo}/{user_login} does not exist: {ex}")
+            logger.debug(f"Project {namespace}/{self.repo} does not exist: {ex}")
         return None
+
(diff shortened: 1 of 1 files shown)


mfocko (approved):
LGTM

I'm not really sure how it is supposed to work generally for users, cause you cannot have `‹username›/forks` for users, `‹username›` must be a group itself. Also user would need the privileges to create repositories in that namespace, right? Or am I missing something?

Also
1. please add release notes to the description (just copy-paste this, till I distribute template)

       ---
       
2. I'm surprised tests are fine even with additional API calls o.o (or maybe it got caught in the exception :thinking:)

jpopelka:
> Or am I missing something?

No, I'm the one who's wrong here.

I thought the user we have for our stream service is [packit-service](https://gitlab.com/packit-service), but it's [packit-as-a-service](https://gitlab.com/packit-as-a-service) and [packit-service](https://gitlab.com/packit-service) is just a group the user is in.
So the service would try to create a fork in "packit-as-a-service/forks", not "packit-service/forks".

This not gonna fly, but I'll keep it open for some time (if you don't mind) and try to think about it.

mfocko:
> This not gonna fly, but I'll keep it open for some time (if you don't mind) and try to think about it.

@jpopelka now that i think about it, it can't work at all, cause on gitlab the user can **never** have subgroup under his username; I don't know where exactly you want to utilize it, but… it's a bit spread, so I would probably go with something like `GitlabForkManager` that would take list of namespaces that are prioritized to forking in `‹username›` namespace:

- internally it could work fine with default constructor taking no prioritized namespaces
- backward compatible
- you wouldn't need to construct the same list (or just one namespace) over and over
- no copy-paste in `fork_create` for different handling ⇒ for-loop + one exception after loop

However I admit it is a bit breaking for the API, since it's becoming fragmented.

PR #632, opened 2021-08-19
Update compatibility.md: remove fully supported properties

nikromen:
`Project.get_files`, `Project.create_issue` with labels and `PullRequest.patch` are now fully supported properties/methods and they should be removed from the `compatibility.md` file.


lbarcziova (approved):
thanks! maybe you can even mention this file in `CONTRIBUTING.md` so that contributors can update it when implementing some missing functionality from here

PR #633, opened 2021-08-19
Let contributors know about changing the COMPATIBILITY.md

nikromen:
Let contributors know about propagate their chandes in our COMPATIBILITY.md file if necessary. 

Maybe we can think about a PR template to remind contributors this as @lachmanfrantisek suggested.


csomh (approved):
:+1:

PR #634, opened 2021-08-19
Release 0.29.0

mmuzila:


lbarcziova (commented):
since these were just changes to markdown files, I think it is not needed to do a release this sprint

mmuzila:
> since these were just changes to markdown files, I think it is not needed to do a release this sprint

ok, then I'm not going to release it.

Issue #199, opened 2019-09-13
Do not let the external exception go to the user


lachmanfrantisek:
It is sometimes hard to get the right problem when a user receives a long traceback with some GitHub/GitLab exception at the end. Those exceptions are not always clear or too general (`e.g. Object not found`).

We can probably do better and raise our exceptions that will have much clearer message since we usually know the context (e.g. `PR with id={id} not found.`).

AC:

Go through the code and cover some parts of code in the try blocks and raise our exception with good message:

```python
    try:
        problematic_code()
    except TheExternalException as ex:
        raise SomeOgrException("Good reason", ex)
```
```

lachmanfrantisek:
It's maybe better after the big code restructure, but we still need to make sure it's done.

mfocko:
as discussed on arch-meeting @ 2021/07/20 this issue should be extended of having generic ogr exception for network (e.g. `ConnectionError` or retries) or internal errors on API endpoints (`50x` return codes). These exceptions are to be processed by `packit-service` or Sentry.

Issue #641, opened 2021-09-27
Improve handling of authentication exceptions


FrNecas:
Follow-up of https://github.com/packit/ogr/issues/199 . If a wrong token is provided (or the token expires), `BadCredentialsException` is thrown from Github. This exception can however pop up almost anywhere in the code so catching it in all these places would be very impractical. There may be a better way how to handle this.

PR #642, opened 2021-09-29
Frnecas exceptions

FrNecas:
TODO:

- [ ] Add test coverage for the newly raised exceptions

Fixes https://github.com/packit/ogr/issues/199 and https://github.com/packit/ogr/issues/641 (didn't want to fix the latter initially but it ended up being not so complicated)

Related to

Merge before/after

---

Ogr now catches internal exceptions from Gitlab and Github and converts them to ogr exceptions, `GitlabAPIException` and `GithubAPIException`, respectively. A new exception, `NetworkError` has been introduced for signalling situations where a request could not be performed due to a network outage.


FrNecas:
Test coverage added, I hope I also resolved the issue with CI failing. I also wanted to somehow mock network failure (and catching the new `NetworkError`) though I am not sure what's the best way to do this

FrNecas:
Tests are passing now but the decorator is now not applied to static and class methods. I will have to play around with these two special cases a bit more.

ogr/services/github/project.py
```diff
@@ -223,7 +223,7 @@ def add_user(self, user: str, access_level: AccessLevel) -> None:
                 user, permission=access_dict[access_level]
             )
         except Exception as ex:
-            raise GithubAPIException("User {user} not found") from ex
+            raise GithubAPIException(f"User {user} not found") from ex
```
mfocko, line 226: :+1:

mfocko (commented):
LGTM so far… do you want to finish the static/class-methods here or in a separate PR?

FrNecas:
I'd like to finish it here. I've made some adjustments to the metaclass, lets see if it works now (also a bit more docs + type hints added).

ogr/abstract.py
```diff
@@ -42,7 +63,67 @@ def cached_property(func):  # type: ignore
...
+            # be used last, especially prior to Python 3.10 since they return
+            # descriptor objects and not functions.
+            # See: https://bugs.python.org/issue43682
+            if isinstance(value, staticmethod):
+                namespace[key] = staticmethod(catch_common_exceptions(value.__func__))
+            elif isinstance(value, classmethod):
+                namespace[key] = classmethod(catch_common_exceptions(value.__func__))
+            elif callable(namespace[key]):
+                namespace[key] = catch_common_exceptions(namespace[key])
+        return super().__new__(cls, name, bases, namespace)
```
FrNecas, line 123: @mfocko Tests are passing now both locally (py3.9) and in the check (py3.10). I've had to tweak this but this implementation seems to be compatible with both versions. I've also added a test for checking authentication failure in a staticmethod. If this looks OK to you, I think we can merge.

lachmanfrantisek:
Looks like we need another rebase, @FrNecas can you please do that?

PR #655, opened 2021-11-12
Add app property to GithubCheckRun

jpopelka:
I'm about to [remove the app name from the check run names](https://github.com/packit/packit-service/pull/1281) so this will be useful (for example in the validation script).

---

N/A

### DIFF
--- a/ogr/services/github/check_run.py
+++ b/ogr/services/github/check_run.py
@@ -6,6 +6,7 @@ from enum import Enum
 from typing import Any, Dict, List, Optional, Union
 
 from github.CheckRun import CheckRun
+from github.GithubApp import GithubApp
 from github.GithubObject import NotSet
 
 from ogr.abstract import OgrAbstractClass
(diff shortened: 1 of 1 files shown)


ogr/services/github/check_run.py
```diff
@@ -168,14 +171,19 @@ def completed_at(self) -> Optional[datetime.datetime]:
         return self.raw_check_run.completed_at
 
     @property
-    def output(self) -> GithubCheckRunOutput:
+    def output(self) -> CheckRunOutput:
```
mfocko, line 174: what was there was just type alias, so it should be fine, not even sure whether the thing provided by PyGithub is some custom data type _that can be used as dictionary_ or just the dictionary :thinking:

ogr/services/github/check_run.py
```diff
@@ -168,14 +171,19 @@ def completed_at(self) -> Optional[datetime.datetime]:
...
         """Output of the check run."""
         return self.raw_check_run.output
 
     @output.setter
     def output(self, output: GithubCheckRunOutput) -> None:
         self.raw_check_run.edit(output=output)
 
+    @property
+    def app(self) -> GithubApp:
```
mfocko, line 183: Oh, I see… K then, if you want to be consistent

mfocko (approved):

PR #640, opened 2021-09-24
Get PR and Issue comment by its ID

nikromen:
I implemented this feature because my PR [packit-service/1222](https://github.com/packit/packit-service/pull/1222) needs to get a comment by its ID but OGR doesn't support that.

This `get_comment` method just simply returns a comment which is identified by comment's ID. Works for PullRequest (not Pagure) and Issue as well.

---

Added support for getting issue and pull request comments by their ID. Pagure currently does not support getting pull request comments.

PR #659, opened 2021-11-30
Add PagurePullRequest.get_comment by its id

nikromen:
I want to implement this even Pagure API doesn't support it because this will solve annoying thing in packit-service for me - we have `comment_object` for github and gitlab and pagure but not for pagure PR :see_no_evil: 

Related to #640 

(i am uncertain if the release note below is useful for users but I'll write it down anyway just to be sure)

---

OGR now fully supports getting PR comments by its id.


lbarcziova (approved):
lgtm

lachmanfrantisek:
> (i am uncertain if the release note below is useful for users but I'll write it down anyway just to be sure)

A user of OGR is a Python developer so it makes sense to have it more technical.

PR #668, opened 2022-01-03
Implement getter of SHA for branches

mfocko:
Required for passing commit hashes from service to testing farm for
merging of tests that are run on their side or usage in tmt preparation.

TODO:

- [x] tests
- [x] consider shared method `get_sha_from(branch=None, tag=None) -> Optional[str]`

---

We have introduced a new function into `ogr` that allows you to get commit SHA of the HEAD of the branch.

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -1708,6 +1708,13 @@ class GitProject(OgrAbstractClass):
         """
         raise NotImplementedError()
 
+    def get_sha_from_branch(self, branch: str) -> Optional[str]:
+        """
+        Returns:
+            Commit SHA of head of the branch. `None` if no branch was found.
+        """
+        raise NotImplementedError()
+
 
 class GitUser(OgrAbstractClass):
     """
--- a/ogr/services/github/project.py
+++ b/ogr/services/github/project.py
@@ -574,3 +574,11 @@ class GithubProject(BaseGitProject):
 
     def get_tags(self) -> List["GitTag"]:
         return [GitTag(tag.name, tag.commit.sha) for tag in self.github_repo.get_tags()]
+
+    def get_sha_from_branch(self, branch: str) -> Optional[str]:
+        try:
+            return self.github_repo.get_branch(branch).commit.sha
+        except GithubException as ex:
+            if ex.status == 404:
+                return None
+            raise GithubAPIException from ex
--- a/ogr/services/gitlab/project.py
+++ b/ogr/services/gitlab/project.py
@@ -5,6 +5,7 @@ import logging
 from typing import List, Optional, Dict, Set, Union
 
 import gitlab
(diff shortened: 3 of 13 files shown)


mfocko:
@csomh @TomasTomecek @lachmanfrantisek I would also like some opinion on the `TODO` and that is changing interface to: `get_sha_from(branch=None, tag=None)` which seems (at least to me) like a „Python-like“ API, but the implementation looks much worse:

```py
# BaseProject
def get_sha_from(branch: str = None, tag: str = None) -> Optional[str]:
  # check if provided «exactly» one of ‹branch› or ‹tag›
  #   throw exception if not
  return None

# projects
def get_sha_from(branch: str = None, tag: str = None) -> Optional[str]:
  super().get_sha_from(branch=branch, tag=tag)

  if branch:
    # code for SHA of branch
    pass

  # code for SHA of tag
  return None
```

On the bright side, I like the idea of only one function, doing only one thing (that is getting the SHA, regardless whether of tag or branch). On the other side, I don't like the way it has to be implemented, SRP of it and I also feel like factoring out getting SHA of branch and tag, which basically means we could end up with 3 functions:

```
get_sha_from(branch=None, tag=None) # this one calls internally the other 2
get_sha_from_tag(tag)
get_sha_from_branch(branch)
```
↑ this also changes the TODO from «changing» to «extending» interface

And my judgement might be also affected by „indentation hell“, because both Git(Hub|Lab) throw an exception if the branch doesn't exist, so it looks meh even without being nested in another block…

csomh:
To me this seems like a syntactic nuance from a user point of view:

```python
get_sha_from(branch="branch")
get_sha_from_branch("branch")
get_sha_from(tag="tag")
get_sha_from_tag("tag")
```

And I don't have the "more Pythonic"-feeling towards any of the above :-)

As you said `get_sha_from()` would complicate implementation (and now that I think about it, could be somewhat confusing for users: it's not obvious just by looking at the method signature if providing no or both arguments is a valid option or not), so I would argue against going that way.

jpopelka (approved):

ogr/abstract.py
```diff
@@ -1708,6 +1708,14 @@ def get_web_url(self) -> str:
         """
         raise NotImplementedError()
 
+    # TODO: Consider ‹get_sha_from(branch=None, tag=None)›
+    def get_sha_from_branch(self, branch: str) -> Optional[str]:
```
TomasTomecek, line 1712: I'd say having two separate methods for each is a better design :) `get_sha_from` sounds like it would be too complicated and prone to errors - also with modern editors you are immediately being offered a list methods so it's easier to pick one which has narrow scope

TomasTomecek (approved):

mfocko:
removed `TODO` from the code

PR #669, opened 2022-01-04
0.34.0 release

TomasTomecek:
--
Hello 2022


fedora/python-ogr.spec
```diff
@@ -52,6 +52,9 @@ rm -rf %{srcname}.egg-info
 
 
 %changelog
+* Tue Jan 04 2022 Tomas Tomecek <ttomecek@redhat.com> - 0.34.0-1
+- 0.34.0 upstream release
```
mfocko, line 56:
```suggestion
- New upstream release 0.34.0
```

pita, but to be consistent

mfocko (approved):

fedora/python-ogr.spec
TomasTomecek: I need to also change that in my shell history since I just rerun the command :D

mfocko: well… we have it like this here, and in `packit` it's the other way :D

PR #670, opened 2022-01-04
Releases rework

### LINKED ISSUE #86: object-specific methods bound to ogr classes

### LINKED ISSUE #112: Support releases in Pagure
Figure out how to handle releases in Pagure.

mfocko:
breaking changes:
- reordered and renamed parameters for `.create_release`
  * also introduced into `GitProject`s API

Fixes #86
Fixes #112

---

RELEASE NOTES BEGIN
Release class in ogr has been reworked and `create_release` has been made part of the API for `GitProject`.
RELEASE NOTES END

### DIFF
--- a/ogr/services/github/release.py
+++ b/ogr/services/github/release.py
@@ -1,35 +1,116 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
+import datetime
+from typing import List, Optional
+from github import GithubException
 from github.GitRelease import GitRelease as PyGithubRelease
 
 from ogr.abstract import Release, GitTag
+from ogr.exceptions import GithubAPIException
 from ogr.services import github as ogr_github
 
 
 class GithubRelease(Release):
+    _raw_release: PyGithubRelease
     project: "ogr_github.GithubProject"
 
-    def __init__(
-        self,
-        tag_name: str,
-        url: str,
-        created_at: str,
-        tarball_url: str,
-        git_tag: GitTag,
-        project: "ogr_github.GithubProject",
-        raw_release: PyGithubRelease,
-    ) -> None:
-        super().__init__(tag_name, url, created_at, tarball_url, git_tag, project)
-        self.raw_release = raw_release
+    @staticmethod
+    def _release_id_from_name(
+        project: "ogr_github.GithubProject", name: str
+    ) -> Optional[int]:
+        releases = project.github_repo.get_releases()
+        for release in releases:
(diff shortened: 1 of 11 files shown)


jpopelka (approved):
Looks OK to me

mfocko:
waiting for #671, since it also contains fixes for new `python-gitlab`

mfocko:
Not sure how much we can rely on the rev-dep tests, since locally (in toolbox) service tests are failing, but here it's green… :confused:

ogr/services/github/release.py
```diff
@@ -1,35 +1,116 @@
...
+        name: str,
+        message: str,
+        ref: Optional[str] = None,
+    ) -> "Release":
+        created_release = project.github_repo.create_git_release(
+            tag=tag, name=name, message=message
+        )
+        return GithubRelease(created_release, project)
+        # XXX: Why?
+        return GithubRelease.get(project, created_release.id)
```
lachmanfrantisek, line 113: Some temporary code change?

ogr/services/gitlab/release.py
```diff
@@ -1,36 +1,89 @@
...
+        return self.project._git_tag_from_tag_name(self.tag_name)
+
+    @property
+    def tag_name(self) -> str:
+        return self._raw_release.tag_name
+
+    @property
+    def url(self) -> Optional[str]:
+        # no link is provided
+        return None
```
lachmanfrantisek, line 37: Can't we guess/construct that URL?

lachmanfrantisek (commented):
Wow, well done!

Thanks for working on this!

ogr/services/github/release.py
mfocko: It was there before and I do not really understand why the ID was needed.

ogr/services/gitlab/release.py
mfocko: Well, now I am puzzled, cause it appears that GitLab has a specific interface for releases and you could potentially guess URL, cause it uses _tag name_ in the URL pointing to the release. Will need to have a better look at it, cause I have no clue how recent that feature is.

PR #672, opened 2022-01-06
Switch ‹Retry› constructor to ‹allowed_methods›

mfocko:
Since ‹method_whitelist› is being deprecated.

**Merge once it doesn't break F34**

---

_N/A_

### DIFF
--- a/ogr/services/github/service.py
+++ b/ogr/services/github/service.py
@@ -65,7 +65,7 @@ class GithubService(BaseGitService):
                 total=int(max_retries),
                 read=0,
                 # Retry mechanism active for these HTTP methods:
-                method_whitelist=["DELETE", "GET", "PATCH", "POST", "PUT"],
+                allowed_methods=["DELETE", "GET", "PATCH", "POST", "PUT"],
                 # Only retry on following HTTP status codes
                 status_forcelist=[500, 503, 403, 401],
                 raise_on_status=False,


jpopelka (approved):
> Merge once it doesn't break F34

and when exactly is that?

mfocko:
> > Merge once it doesn't break F34
> 
> and when exactly is that?

I wonder that too… Especially since it's fine on el8 :D it's coming from `urllib` and that is „standard library“, so it might be python version too :thinking:

mfocko:
that's even more funny:

| target | python version | state  |
| ------ | -------------- | ------ |
| el8    | 3.6.8          | `PASS` |
| f34    | 3.9.9          | `FAIL` |
| f35    | 3.10.1         | `PASS` |

FrNecas:
/packit test

.. TF runs no longer available, lets see if this is still an issue

mfocko:
> /packit test
> 
> .. TF runs no longer available, lets see if this is still an issue

Doesn't work on F34 (as can be seen in comments), this will more than likely wait for the decommission :/

FrNecas:
Maybe in the meantime we could use try-except just so that we are not throwing warnings and can merge stuff with green tests? What do you think?

mfocko:
> Maybe in the meantime we could use try-except just so that we are not throwing warnings and can merge stuff with green tests? What do you think?

I will have a look into that on Friday, `__version__` could be an option too. :thinking:

PR #679, opened 2022-02-15
Deprecation warning with Github Retry kwarg

NikolasKomonen:
We were getting a warning during our pytest tests:
```
DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
    self._max_retries = Retry(
```
Changelog here: https://github.com/urllib3/urllib3/blob/9571a9860437f4ce2ebb04803a90625db5ba52ee/CHANGES.rst#1260-2020-11-10

This should hopefully resolve it, not sure if there are any tests that need to be updated regarding this.

### DIFF
--- a/ogr/services/github/service.py
+++ b/ogr/services/github/service.py
@@ -65,7 +65,7 @@ class GithubService(BaseGitService):
                 total=int(max_retries),
                 read=0,
                 # Retry mechanism active for these HTTP methods:
-                method_whitelist=["DELETE", "GET", "PATCH", "POST", "PUT"],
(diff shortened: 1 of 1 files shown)


mfocko:
Hi, thanks for contributing to Packit. Unfortunately it is a duplicate of #672, which is being held unmerged for now because there is an issue with switching to `allowlist` on Fedora 34.

NikolasKomonen:
@mfocko sorry didn't see that. Thanks, I'll close this then.

PR #707, opened 2022-04-28
Release 0.38.0

FrNecas:


mfocko (approved):
Not sure if we want to wait for switching to stage on downstream or not.

lachmanfrantisek (approved):
:+1:

Regarding Packit, we agreed on using prod&stg for upstream including `propose-downstream` and just `stg` for downstream (Koji and Bodhi). And this is actually now configured. So we are good to go..

PR #715, opened 2022-05-19
When merging from fork the target is the fork's parent

majamassarini:
The PullRequest target project should be the parent of the forked project when merging from fork.

Fixes packit/hardly#63

### DIFF
--- a/tests/integration/github/test_pull_requests.py
+++ b/tests/integration/github/test_pull_requests.py
@@ -87,6 +87,7 @@ class PullRequests(GithubTests):
 
         assert pr_upstream_upstream.title == "test: upstream <- upstream"
         assert pr_upstream_upstream.status == PRStatus.open
+        assert not pr_upstream_upstream.target_project.is_fork
         assert pr_opened_after == pr_opened_before + 1
 
         pr_upstream_upstream.close()
@@ -117,6 +118,7 @@ class PullRequests(GithubTests):
             == "test: upstream <- fork_username:source_branch"
         )
         assert pr_upstream_forkusername.status == PRStatus.open
+        assert not pr_upstream_forkusername.target_project.is_fork
         assert pr_opened_after == pr_opened_before + 1
 
         pr_upstream_forkusername.close()
@@ -142,6 +144,7 @@ class PullRequests(GithubTests):
 
         assert pr_upstream_fork.title == "test: upstream <- fork"
         assert pr_upstream_fork.status == PRStatus.open
+        assert not pr_upstream_fork.target_project.is_fork
         assert pr_opened_after == pr_opened_before + 1
 
         pr_upstream_fork.close()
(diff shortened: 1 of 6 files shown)


mfocko (approved):
LGTM :) could you please update the same thing in the GitHub and GitLab implementation?

As I have mentioned in the mail, it should be those lines:
GitLab: https://github.com/packit/ogr/blob/3fd707590a3545109f3ce10e14e96c89f200a9b9/ogr/services/gitlab/pull_request.py#L197
GitHub: https://github.com/packit/ogr/blob/3fd707590a3545109f3ce10e14e96c89f200a9b9/ogr/services/github/pull_request.py#L165

Ideally I would set a variable with `target_project` ot be explicit about it when passing the project to the constructor

majamassarini:
> LGTM :) could you please update the same thing in the GitHub and GitLab implementation?

Yes, sure. I think the hardest thing is to build a test because I have no way to trigger the error in gitlab or github. I'll try.

mfocko:
> > LGTM :) could you please update the same thing in the GitHub and GitLab implementation?
> 
> Yes, sure. I think the hardest thing is to build a test because I have no way to trigger the error in gitlab or github. I'll try.

If you want to have failed tests, it should be enough to:
1. fork some testing repository
2. create PR by calling `.create_pr` on your fork
3. enjoy red tests, cause `target_project` should be your fork instead of the upstream

(at least that's the way you can reproduce it for GitHub and GitLab since you have already fixed the Pagure)

majamassarini:
> > Yes, sure. I think the hardest thing is to build a test because I have no way to trigger the error in gitlab or github. I'll try.
> 
> If you want to have failed tests, it should be enough to:
> 
> 1. fork some testing repository
> 2. create PR by calling `.create_pr` on your fork
> 3. enjoy red tests, cause `target_project` should be your fork instead of the upstream
> 
> (at least that's the way you can reproduce it for GitHub and GitLab since you have already fixed the Pagure)

Nice, Thanks!

tests/integration/github/test_pull_requests.py
```diff
@@ -87,6 +87,7 @@ def test_pr_create_upstream_upstream(self):
 
         assert pr_upstream_upstream.title == "test: upstream <- upstream"
         assert pr_upstream_upstream.status == PRStatus.open
+        assert pr_upstream_upstream.target_project.is_fork == False
```
mfocko, line 90:
```suggestion
        assert not pr_upstream_upstream.target_project.is_fork
```

also flake from pre-commit doesn't like it the way it is now ;)

mfocko (approved):
LGTM :)

tests/integration/github/test_pull_requests.py
majamassarini: 😅

mfocko (approved):

jpopelka:
Tested on Gitlab (staging CentOS Stream source-git/dist-git) and looks OK  :rocket:

PR #723, opened 2022-07-14
Enforce strict Optional typing

### LINKED ISSUE #696: Run mypy without --no-strict-optional
1. remove [`--no-strict-optional` from pre-commit-config](https://github.com/packit/ogr/blob/1e7d9320c9df1a10aecce815f339c3b85c7471e5/.pre-commit-config.yaml#L41)
2. run `pre-commit run --all-files`
3. fix all errors

https://mypy.readthedocs.io/en/stable/kinds_of_types.html#no-strict-optional

FrNecas:
Some of these fixes are not perfect, in a few places I added `Optional` where I feel it shouldn't really be. However, it's not really possible any other way without significant refactoring which would break the API (which I don't want to do). Sometimes a specific forge may return `None` so it was necessary to update the type of the abstract class.

I tried to not make any functional changes, most of these are just to align the typing with the actual implementation.

TODO:

- [ ] regenerate some requre responses (the failing tests). I may need some help with this, I managed to do this for one of the failing tests but not for the repo creation yet.

Fixes #696 

Related to #251 

Merge before/after

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -217,7 +217,7 @@ class Comment(OgrAbstractClass):
         self._body = new_body
 
     @property
-    def id(self) -> int:
+    def id(self) -> Optional[int]:
         return self._id
 
     @property
@@ -226,12 +226,12 @@ class Comment(OgrAbstractClass):
         return self._author
 
     @property
-    def created(self) -> datetime.datetime:
+    def created(self) -> Optional[datetime.datetime]:
         """Datetime of creation of the comment."""
         return self._created
 
     @property
-    def edited(self) -> datetime.datetime:
+    def edited(self) -> Optional[datetime.datetime]:
         """Datetime of last edit of the comment."""
         return self._edited
 
@@ -256,7 +256,7 @@ class Comment(OgrAbstractClass):
 
 class IssueComment(Comment):
     @property
-    def issue(self) -> "Issue":
+    def issue(self) -> Optional["Issue"]:
         """Issue of issue comment."""
         return self._parent
 
@@ -266,7 +266,7 @@ class IssueComment(Comment):
 
 class PRComment(Comment):
     @property
-    def pull_request(self) -> "PullRequest":
+    def pull_request(self) -> Optional["PullRequest"]:
         """Pull request of pull request comment."""
         return self._parent
 
@@ -642,12 +642,12 @@ class PullRequest(OgrAbstractClass):
         raise NotImplementedError()
 
     @property
-    def target_branch_head_commit(self) -> str:
+    def target_branch_head_commit(self) -> Optional[str]:
         """Commit hash of the HEAD commit of the target branch."""
         raise NotImplementedError()
 
     @property
-    def merge_commit_sha(self) -> str:
+    def merge_commit_sha(self) -> Optional[str]:
         """
         Commit hash of the merge commit of the pull request.
 
@@ -1009,12 +1009,12 @@ class CommitFlag(OgrAbstractClass):
         raise NotImplementedError()
 
     @property
-    def created(self) -> datetime.datetime:
+    def created(self) -> Optional[datetime.datetime]:
         """Datetime of creating the commit status."""
         raise NotImplementedError()
 
     @property
-    def edited(self) -> datetime.datetime:
+    def edited(self) -> Optional[datetime.datetime]:
         """Datetime of editing the commit status."""
         raise NotImplementedError()
 
@@ -1230,7 +1230,7 @@ class GitService(OgrAbstractClass):
         instance_url (str): URL of the git forge instance.
     """
 
-    instance_url: Optional[str] = None
+    instance_url: str
 
     def __init__(self, **_: Any) -> None:
         pass
@@ -1326,7 +1326,9 @@ class GitService(OgrAbstractClass):
 
 
 class GitProject(OgrAbstractClass):
-    def __init__(self, repo: str, service: GitService, namespace: str) -> None:
+    def __init__(
+        self, repo: str, service: GitService, namespace: Optional[str]
+    ) -> None:
         """
         Args:
             repo: Name of the project.
@@ -1335,7 +1337,7 @@ class GitProject(OgrAbstractClass):
 
                 - GitHub: username or org name.
                 - GitLab: username or org name.
-                - Pagure: namespace (e.g. `"rpms"`).
+                - Pagure: namespace (e.g. `"rpms"`). May be none.
 
                   In case of forks: `"fork/{username}/{namespace}"`.
         """
@@ -1916,7 +1918,7 @@ class GitUser(OgrAbstractClass):
         """
         raise NotImplementedError()
 
-    def get_email(self) -> str:
+    def get_email(self) -> Optional[str]:
         """
         Returns:
             Email of the user.
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -1009,7 +1009,7 @@ class CommitFlag(OgrAbstractClass):
         raise NotImplementedError()
 
     @property
-    def created(self) -> Optional[datetime.datetime]:
+    def created(self) -> datetime.datetime:
         """Datetime of creating the commit status."""
         raise NotImplementedError()
 
@@ -1112,7 +1112,11 @@ class Release(OgrAbstractClass):
 
     @property
     def git_tag(self) -> GitTag:
-        """Object that represents tag tied to the release."""
+        """Object that represents tag tied to the release.
+
+        Raises:
+            OgrException, if the tag is not found.
+        """
         raise NotImplementedError()
 
     @property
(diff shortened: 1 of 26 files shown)


jpopelka (commented):
> Some of these fixes are not perfect

I see, it's been clear to me that the --no-strict-optional removal would mean some compromises.

> TODO: regenerate some requre responses (the failing tests). I may need some help with this,

I'll try, but don't promise anything (I usually try to avoid this :-) )

mfocko:
I'll have a look.

BTW it kinda bothers me that you need to regenerate data after typing-related changes :D

ogr/services/pagure/service.py
```diff
@@ -332,6 +332,8 @@ def project_create(
             parameters["description"] = repo
         if namespace:
             parameters["namespace"] = namespace
+        else:
+            namespace = self.user.get_username()
```
FrNecas, line 336: @mfocko in some cases, changing the code rather than the typing just seemed like the more correct solution. This is one of the places which causes the tests to fail. If `namespace` is not provided, it seems to assume the current authenticated user and fills it in automatically. However, `PagureProject` constructor requires namespace. If I hadn't added this extra call (which obviously creates the need for test data regeneration), the only solution would be to make `namespace` inside `PagureProject` `Optional` which does not seem right.

ogr/abstract.py
```diff
@@ -217,7 +217,7 @@ def body(self, new_body: str) -> None:
         self._body = new_body
 
     @property
-    def id(self) -> int:
+    def id(self) -> Optional[int]:
```
mfocko, line 220: Does it make sense though?

I think I know what you meant in the description.

ogr/abstract.py
```diff
@@ -1335,7 +1337,7 @@ def __init__(self, repo: str, service: GitService, namespace: str) -> None:
 
                 - GitHub: username or org name.
                 - GitLab: username or org name.
-                - Pagure: namespace (e.g. `"rpms"`).
+                - Pagure: namespace (e.g. `"rpms"`). May be none.
```
mfocko, line 1340:
```suggestion
                - Pagure: namespace (e.g. `"rpms"`). May be `None`, i.e. no namespace present.
```

ogr/parsing.py
```diff
@@ -10,7 +10,7 @@ class RepoUrl:
     Class that represents repo URL.
 
     Attributes:
-        repo (str): Name of the repository.
+        repo (Optional[str]): Name of the repository. Can be None if parsing is being done.
```
mfocko, line 13:
```suggestion
        repo: Name of the repository. Can be `None` if parsing is being done.
```

```suggestion
        repo (Optional[str]): Name of the repository. Can be `None` if parsing is being done.
```

ogr was the first project with the google-style docstrings, in the end we have decided to not keep the trailing types after attributes. I guess, it's up to you, if we start to slowly remove them, where we touch the code or not.

ogr/services/github/auth_providers/token.py
```diff
@@ -10,7 +10,9 @@
 
 
 class TokenAuthentication(GithubAuthentication):
-    def __init__(self, token: str, max_retries: Union[int, Retry] = 0, **_) -> None:
+    def __init__(
+        self, token: Optional[str], max_retries: Union[int, Retry] = 0, **_
```
mfocko, line 14: oh no, memories of this magic are coming back /o\

ogr/services/github/flag.py
```diff
@@ -64,9 +64,13 @@ def set(
         return GithubCommitFlag(project=project, raw_commit_flag=status, commit=commit)
 
     @property
-    def created(self) -> datetime.datetime:
-        return self._raw_commit_flag.created_at
+    def created(self) -> Optional[datetime.datetime]:
+        if self._raw_commit_flag:
```
mfocko, line 68: I have a weird feeling that this shouldn't happen :thinking: and in case it happens, we should be „angry“

ogr/services/github/pull_request.py
```diff
@@ -148,17 +148,18 @@ def create(
...
-            github_repo = project.parent.github_repo
-            target_project = project.parent
-        elif fork_username:
-            source_branch = f"{fork_username}:{source_branch}"
-            if fork_username != project.namespace and project.parent is not None:
-                github_repo = GithubPullRequest.__get_fork(
-                    fork_username, project.parent.github_repo
-                )
+        if project.parent:
+            if project.is_fork and fork_username is None:
```
mfocko, line 152: this is exactly one thing that I don't like on static typing :D it has no clue about invariants and integrity

ogr/services/pagure/release.py
```diff
@@ -71,6 +71,8 @@ def create(
         message: str,
         ref: Optional[str] = None,
     ) -> "Release":
+        if not ref:
+            raise PagureAPIException("Release ref must be specified")
```
mfocko, line 75: I'm leaning towards `ArgumentError`

ogr/services/github/release.py
```diff
@@ -45,7 +45,10 @@ def body(self):
 
     @property
     def git_tag(self) -> GitTag:
-        return self.project.get_tag_from_tag_name(self.tag_name)
+        tag = self.project.get_tag_from_tag_name(self.tag_name)
+        if not tag:
+            raise OgrException(f"No tag with name '{self.tag_name}' found.")
```
mfocko, line 50: I've noticed you used `Raises:` in docstring above, is this documented?

ogr/services/gitlab/issue.py
```diff
@@ -47,7 +47,7 @@ def url(self) -> str:
         return self._raw_issue.web_url
 
     @property
-    def assignees(self) -> list:
+    def assignees(self) -> Optional[list]:
```
mfocko, line 50:
```suggestion
    def assignees(self) -> Optional[List[???]]:
```

I'm surprised mypy isn't yelling :D

I hope for `s/???/str`, but it's GitLab, so I'm not willing to bet on it :upside_down_face:

ogr/services/gitlab/service.py
```diff
@@ -104,6 +104,8 @@ def project_create(
             except gitlab.GitlabGetError as ex:
                 raise GitlabAPIException(f"Group {namespace} not found.") from ex
             data["namespace_id"] = group.id
+        else:
+            namespace = self.user.get_username()
```
mfocko, line 108: this one is breaking tests, right?

ogr/services/pagure/pull_request.py
```diff
@@ -150,10 +150,11 @@ def create(
         }
 
         caller = project
-        if project.is_fork:
+        if project.is_fork and caller.parent:
```
mfocko, line 153: tbf, this could break tests too, because you need to query both `is_fork` and `parent`, and on the background it may be 2 separate API calls, even though one is enough :/

mfocko (commented):
yup, some of those `Optional`s are quite suspicious

mfocko:
Oh, only 4 tests? I expected worse :grin:

ogr/abstract.py
FrNecas: Yes, this is a bit of a dilemma. It doesn't make much sense semantically, however just based on our interface, this could happen. If I wanted to enforce an `int` here, the order of arguments in the constructor would have to be changed (along with some other changes of course)

ogr/services/github/flag.py
FrNecas: Hm, good point, let's raise an exception instead of `None` in case `_raw_commit_flag` is not set.

ogr/services/pagure/pull_request.py
FrNecas: Yes, this breaks something too IIRC.

ogr/services/github/pull_request.py
FrNecas: You are right, if it is a fork, then it definitely does have a parent but mypy doesn't know anything about it. I am not sure if we can somehow encode this relationship (I don't think so). One more check doesn't hurt and is safer... except that in some cases requre will shout :D

FrNecas:
I've resolved some of the issues pointed out in the review, however I am still a bit unsure about the correct solution in some of the weird cases.

ogr/abstract.py
```diff
@@ -1335,7 +1341,7 @@ def __init__(self, repo: str, service: GitService, namespace: str) -> None:
 
                 - GitHub: username or org name.
                 - GitLab: username or org name.
-                - Pagure: namespace (e.g. `"rpms"`).
+                - Pagure: namespace (e.g. `"rpms"`). May be `None`, i.e. no namespace present.
```
lachmanfrantisek, line 1344: What about using an empty string for this?

(I am looking at that from the `GitLab` perspecive where you can have `"one_namespace"`, `"two/namespaces"`, so why not `""`.)

I know, it's a functional change...;)

lachmanfrantisek (commented):
Wow, thanks so much for all the work!

mfocko:
> I've resolved some of the issues pointed out in the review, however I am still a bit unsure about the correct solution in some of the weird cases.

I think one of the issues with types is that we were using some of the »abstract« classes in the test cases all over the codebase of packit(-service), which makes the typing more complicated.

IMO in ideal case we should have everything hidden behind a property, so that we can handle it from specific git-forge and maybe have a „dummy“ classes for tests. However that will involve breaking of a lot of rev-dep tests and seems, at least to me, time-consuming to fix everywhere :/

lbarcziova:
Closing as this got outdated and hasn't been worked on for years.

PR #729, opened 2022-08-31
Enable EPEL 9

nforro:


nforro:
I didn't realize ogr has runtime dependency on `python-gitlab`, but that's already in [testing](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2022-d9ca6d6d8c), so this shouldn't be blocked for long.

nforro:
recheck

lachmanfrantisek (approved):

PR #759, opened 2022-12-15
Add VM image build configuration for fedora 36

majamassarini:
I would like to use this PR to create a demo for vm image build integration in packit-service.


.idea/inspectionProfiles/profiles_settings.xml
```diff
@@ -0,0 +1,6 @@
+<component name="InspectionProjectProfileManager">
+  <settings>
+    <option name="USE_PROJECT_PROFILE" value="false" />
+    <version value="1.0" />
+  </settings>
+</component>
```
lachmanfrantisek, line 6: I know it's just for the demo, but you've probably submitted the idea files by mistake...;)

.packit.yaml
```diff
@@ -103,3 +103,25 @@ jobs:
       - fedora-latest # branched version, rawhide updates are created automatically
       - fedora-stable
       - epel-9
+
+  # vm image build
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      - fedora-all
+    project: vm-image-ogr-demo
```
lachmanfrantisek, line 112: I am worried this will collide with the statuses from the other `copr_build`. (If it is just for the demo, you can delete the original one.)

.packit.yaml
```diff
@@ -103,3 +103,26 @@ jobs:
...
+      - fedora-37-x86_64
+    project: vm-image-ogr-demo
+
+  - job: vm_image_build
+    trigger: pull_request
+    copr_chroot: fedora-36-x86_64
+    owner: packit
+    project: vm-image-ogr-demo
+    image_customizations:
+      packages: [python3-ogr]
```
mfocko, line 121: Do you actually need this, if the image is generated from the Copr build?

majamassarini: Do you mean specify the package? I think yes but @TomasTomecek knows this better than me.

.idea/inspectionProfiles/profiles_settings.xml
majamassarini: Yes 🙈

.packit.yaml
majamassarini: I try to use the `identifier` key

.packit.yaml
mfocko: Yup, the package. Agreed with you :D Though I'd expect this to be automatically installed because of the Copr build

TomasTomecek: Yes we need this right now.

I guess we could default to all binary RPMs produced by the latest build though we don't have a mechanism right now for that. Hence it's explicit like this.

PR #769, opened 2023-02-14
Warnings make packit fail in parsing sandcastle command output

majamassarini:
When we run the following as a sandcastle command:
`python3 setup.py --version`

we got 2 warning messages that make packit fail in **parsing the command output**:

1. `licence_file` is deprecated, which should be resolved by the first commit
2. `Requirements should be satisfied by a PEP 517 installer`:
    when packit/packit#1850 will be closed this should disappear,
    in the meantime just drop the warning, which should be solved by the second commit.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -20,7 +20,7 @@ actions:
   create-archive:
     - python3 setup.py sdist --dist-dir ./fedora/
     - bash -c "ls -1t ./fedora/*.tar.gz | head -n 1"
-  get-current-version: python3 setup.py --version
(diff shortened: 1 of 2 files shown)


TomasTomecek (approved):
the -Wignore is unfortunate but without fixing the packaging properly, we're stuck with it

thank you for resolving this!

majamassarini:
> the -Wignore is unfortunate but without fixing the packaging properly, we're stuck with it

I agree, but I did not dare to solve the packaging issue :) but I have added here packit/packit#1850 a reminder to remove this second commit...

PR #771, opened 2023-02-21
Compare hostnames when using custom instances

lachmanfrantisek:
Fixes an issue with project->service mapping where the service with an url not containing the service type wasn't matched.

Related to

Merge before/after

RELEASE NOTES BEGIN
Fixes an issue with project->service mapping where the service with an url not containing the service type wasn't matched.
RELEASE NOTES END

### DIFF
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -64,7 +64,7 @@ def get_project(
     Args:
         url: URL of the project, e.g. `"https://github.com/packit/ogr"`.
         service_mapping_update: Custom mapping from
-            service url (`str`) to service class.
+            service url/hostname (`str`) to service class.
 
             Defaults to no mapping.
         custom_instances: List of instances that will be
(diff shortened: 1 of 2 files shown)


ogr/factory.py
```diff
@@ -84,7 +84,7 @@ def get_project(
     mapping = service_mapping_update.copy() if service_mapping_update else {}
     custom_instances = custom_instances or []
     for instance in custom_instances:
-        mapping[instance.instance_url] = instance.__class__
+        mapping[instance.hostname] = instance.__class__
```
nforro, line 87: Shouldn't the docstring be updated as well?

nforro (approved):

ogr/factory.py
lachmanfrantisek: Good point, but after checking the docstring, I don't think it's needed. This is more a bugfix than a change.

nforro: I mean, the doc for `service_mapping_update` says `Custom mapping from service url (str) to service class.` and you are updating that with hostname to class mappings.

lachmanfrantisek: Ah, I see... Thanks!

lachmanfrantisek: OK, to preserve old docstring and functionality, I've made the functions to support both hostname and URL.

@nforro PTAL, thanks!

nforro (approved):

PR #784, opened 2023-05-17
Support PyGithub >= 1.58

nforro:


mfocko:
> So, it seems that requre has issues on EL9 when coverage is enabled. Should I disable it until I try to figure out a proper fix?

I'm not sure I follow… You noticed the issue locally?

nforro:
No, for example here (one of the runs before I pushed https://github.com/packit/ogr/pull/784/commits/d29051ba2f17426c0b1096d7d74483af3c60190d): https://artifacts.dev.testing-farm.io/8772b513-7d7e-4ee6-9d15-6aa19fea0990/

PR #800, opened 2023-09-07
Fix get_fork for Pagure

lbarcziova:
When getting the fork from existing forks, check for equality of the usernames and not only substring check with 'in' operator. This prevents getting a fork of 'packit-stg' for user 'packit'.

Fixes packit/packit-service#2178

RELEASE NOTES BEGIN
We have fixed a bug in `get_fork` method for Pagure about checking the usernames for a match when going through existing forks.
RELEASE NOTES END


mfocko (approved):

mfocko:
I guess we could add release notes to this :smile:

PR #804, opened 2023-09-22
Don't use obsolete setuptools_scm_git_archive

hroncok:
See https://github.com/Changaco/setuptools_scm_git_archive/blob/master/README.rst

TODO:

- [ ] see if it works

Related to https://bugzilla.redhat.com/show_bug.cgi?id=2240107

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,1 +1,4 @@
+node: $Format:%H$
+node-date: $Format:%cI$
+describe-name: $Format:%(describe:tags=true)$
 ref-names: $Format:%D$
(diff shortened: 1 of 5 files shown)


hroncok:
I am overwhelmed by the amount of CI here.

Nevertheless, this won't work on EPEL 9. No idea how to handle that.

nforro:
Thanks.

> Nevertheless, this won't work on EPEL 9. No idea how to handle that.

The thing is, git-archive support is only needed for building from a git archive, e.g. the archive that GitHub provides, building from PyPI sdist or git checkout works just fine with `setuptools_scm` < 7 and no `setuptools_scm_git_archive`.

However, there were too many issues so I had to resort to a bigger change: https://github.com/packit/ogr/pull/808

PR #806, opened 2023-09-24
Ruff

mfocko:


mfocko:
recheck

ruff.toml
```diff
@@ -0,0 +1,25 @@
...
+    "RUF", # Ruff-specific rules
+    "SIM", # flake8-simplify
+    "UP",  # pyupgrade
+    "W",   # pycodestyle Warning
+]
+ignore = [
+    "B017",    # assert-raises-exception
+    "B022",    # useless-contextlib-suppress
+    "RUF002",  # you shall not touch my guillemets
+    "RUF003"   # Comment contains ambiguous character
```
jpopelka, line 21: Have you checked whether we still need those that you copied from packit?
Maybe we don't actually need to ignore them here in ogr.

jpopelka (approved):

ruff.toml
mfocko: I'll have a look at it, there is still one issue with “dissecting” the exceptions in tests :/

jpopelka:
I've remembered that that's still #696 and #723 which are also AFAIR about adding more `Optional` to the code (as your last commit does). I don't quite remember the issues why we haven't finished the #723 but can you please double-check that your last commit doesn't somehow hit the same problems?

mfocko:
> I've remembered that that's still #696 and #723 which are also AFAIR about adding more `Optional` to the code (as your last commit does). I don't quite remember the issues why we haven't finished the #723 but can you please double-check that your last commit doesn't somehow hit the same problems?

Had a look, tl;dr:
* `Optional` has been added to some places where it doesn't make much sense, like PR ID was to be wrapped in `Optional` just because of what gets returned from the dependency :/
* failing requre tests…

I'll go through the changes to check it here, but I doubt it either includes the errors or supersedes the PR :/

PR #819, opened 2023-10-19
gitlab list MRs: return all, not just 20

TomasTomecek:
TODO:

- [x] requre

> python-gitlab==3.14.0

```
Calling a `list()` method without specifying `get_all=True` or
`iterator=True` will return a maximum of 20 items. Your query returned
20 of 87 items.
```

RELEASE NOTES BEGIN
Fixed an issue where getting a list of GitLab merge requests using `.list()` would return only 20 items.
RELEASE NOTES END

### DIFF
--- a/ogr/services/gitlab/pull_request.py
+++ b/ogr/services/gitlab/pull_request.py
@@ -240,10 +240,14 @@ class GitlabPullRequest(BasePullRequest):
         status: PRStatus = PRStatus.open,
     ) -> list["PullRequest"]:
         # Gitlab API has status 'opened', not 'open'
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (commented):
I am surprised this isn't covered by any test. Also, I have a deja vu we've hit and fixed this somewhere already..;)

TomasTomecek:
I couldn't reproduce the CI failure locally with requre, let's see

TomasTomecek:
There is a change in how GitLab calls `.list()` in F37 and the rest.

Well, F37 will be EOL *soon*, so...

mfocko (approved):

PR #823, opened 2023-11-02
docs: add an example of syncing labels

mfocko:


lachmanfrantisek (approved):
I have somewhere a code improving the update method to update the text/color as well...  I should really send it as a PR..;)

mfocko:
I think it got synced :thinking:

PR #824, opened 2023-12-04
Fix GithubRelease.url to not return API url

lbarcziova:
Change the property so that it returns the html_url of the raw release and not the url which is the API url.

Merge before/after

RELEASE NOTES BEGIN
We have fixed a bug that GithubRelease.url returned an API URL.
RELEASE NOTES END


mfocko (approved):
It's a bug fix, you could probably add the release notes.

PR #827, opened 2023-12-19
Fix race condition causing ENOPRSTATS

majamassarini:
When we get the ENOPRSTATS exception from Pagure there is a race condition. 
We can try to run the call again and see if it solves the problem.

Fixes packit/packit-service#2287

Merge before packit/packit-service#2289


ogr/services/pagure/pull_request.py
```diff
@@ -191,18 +192,43 @@ def get(project: "ogr_pagure.PagureProject", pr_id: int) -> "PullRequest":
...
+                    )
+                    if attempt <= retries:
+                        logger.error(
+                            f"Trying again; attempt={attempt} after {wait_seconds}seconds",
+                        )
+                        attempt += 1
+                        sleep(wait_seconds)
+                    else:
+                        raise ex
+                raise ex
```
nforro, line 231: I think you are missing `continue` after `sleep()`, otherwise this will always raise.

ogr/services/pagure/pull_request.py
```diff
@@ -191,18 +192,43 @@ def get(project: "ogr_pagure.PagureProject", pr_id: int) -> "PullRequest":
...
-                method="GET",
-            )
-        except PagureAPIException as ex:
-            if "No statistics" in ex.pagure_error:
-                return {}
-            raise ex
+    def get_files_diff(
+        project: "ogr_pagure.PagureProject",
+        pr_id: int,
+        retries=3,
```
nforro, line 198: I would set retries to 0 by default so that if called without optional arguments it behaves like a direct API call.

```suggestion
        retries=0,
```

ogr/services/pagure/pull_request.py
```diff
@@ -191,18 +192,43 @@ def get(project: "ogr_pagure.PagureProject", pr_id: int) -> "PullRequest":
...
+        wait_seconds=3,
+    ) -> dict:
+        """Collect all the changes in the PR and deal with ENOPRSTATS exception.
+        While the PR is changing its status from open to merged
+        exception can be raised by the Pagure service.
+        Wait for a while to see if we can manage to access changes adjusting the
+        retries and wait_seconds parameters.
+
+        """
+        attempt = 0
```
nforro, line 208: I think we should start at the first attempt:

```suggestion
        attempt = 1
```

tests/integration/pagure/test_pull_requests.py
```diff
@@ -125,6 +127,5 @@ def test_pr_diff(self):
         assert "README.md" in diff
 
     def test_pr_diff_empty_diff(self):
-        diff = self.ogr_project.get_pr_files_diff(6)
-        assert isinstance(diff, dict)
-        assert diff == {}
+        with pytest.raises(PagureAPIException) as _:
+            self.ogr_project.get_pr_files_diff(6, retries=1)
```
nforro, line 131:
```suggestion
        with pytest.raises(PagureAPIException):
            self.ogr_project.get_pr_files_diff(6)
```

ogr/services/pagure/pull_request.py
nforro: We need a test case for fails first, succeeds when retried.

majamassarini: > I think you are missing `continue` after `sleep()`, otherwise this will always raise.

then it complains that the else statement is not usefull :) I think it should be fine like it is now

majamassarini: > We need a test case for fails first, succeeds when retried.

We should have both. `test_pr_diff` when the response is 200 and `text_pr_diff_empty_diff` with 400 error

nforro: > then it complains that the else statement is not usefull :) I think it should be fine like it is now

No, it's not, you would have to add another else above this. But it would be better to `continue` after sleep and remove the else it's complaining about.

ogr/services/pagure/pull_request.py
```diff
@@ -191,18 +192,43 @@ def get(project: "ogr_pagure.PagureProject", pr_id: int) -> "PullRequest":
...
+                    method="GET",
+                )
+            except PagureAPIException as ex:  # noqa PERF203
+                if "No statistics" in ex.pagure_error:
+                    # this may be a race condition, try once more
+                    logger.error(
+                        f"While retrieving PR diffstats Pagure returned ENOPRSTATS. \n{ex}",
+                    )
+                    if attempt <= retries:
+                        logger.error(
```
nforro, line 224: I think debug is enough in this case:

```suggestion
                        logger.debug(
```

ogr/services/pagure/pull_request.py
nforro: > > We need a test case for fails first, succeeds when retried.
> 
> We should have both. `test_pr_diff` when the response is 200 and `text_pr_diff_empty_diff` with 400 error

Neither of them tests the complete code and the case I described and this PR is supposed to solve.

nforro: This is the kind of test I had in mind: https://github.com/majamassarini/ogr/pull/1
It also proves my point about the missing `continue` :slightly_smiling_face:

majamassarini: > > then it complains that the else statement is not usefull :) I think it should be fine like it is now
> 
> No, it's not, you would have to add another else above this. But it would be better to `continue` after sleep and remove the else it's complaining about.

I had forgotten an else statement... Now it should be better even without the continue key.

tests/integration/pagure/test_data/test_pull_requests/PullRequests.test_pr_diff_failing_and_succeding.yaml
```diff
@@ -0,0 +1,169 @@
...
+              Python/3.6
+            Set-Cookie: pagure=eyJfcGVybWFuZW50Ijp0cnVlLCJjc3JmX3Rva2VuIjoiMzU0YmExMTNlMWFhMGU0MzliZDFjN2UwZGY3ODg4OTBjODdhMzkwMSJ9.GFnGGw.MvcSSJ61ycB9FCodU3kj9hWZeGY;
+              Expires=Fri, 12-Jan-2024 10:23:23 GMT; Secure; HttpOnly; Path=/
+            Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
+            X-Content-Type-Options: nosniff
+            X-Frame-Options: ALLOW-FROM https://pagure.io/
+            X-Xss-Protection: 1; mode=block
+          raw: !!binary ""
+          reason: BAD REQUEST
+          status_code: 400
```
nforro, line 116: Here the first request succeeds and the second fails, that doesn't really make sense, they should be swapped. I think it will make the test fail too because the second request is extra.

ogr/services/pagure/pull_request.py
nforro: Yes, now the logic is correct, however instead of this:

```python
                        sleep(wait_seconds)
                    else:
                        raise ex
                else:
                    raise ex
```
you could have this:

```python
                        sleep(wait_seconds)
                        continue
                raise ex
```

Don't you think it's better?

tests/integration/pagure/test_data/test_pull_requests/PullRequests.test_pr_diff_failing_and_succeding.yaml
majamassarini: The test should succeed, at least I run it locally and it was ok. But you are right the order is from top to bottom.

ogr/services/pagure/pull_request.py
majamassarini: Now I see your point 😄👍🏻

tests/integration/pagure/test_pull_requests.py
```diff
@@ -124,7 +126,11 @@ def test_pr_diff(self):
...
         assert "README.md" in diff
 
-    def test_pr_diff_empty_diff(self):
-        diff = self.ogr_project.get_pr_files_diff(6)
+    def test_pr_diff_failing(self):
+        with pytest.raises(PagureAPIException):
+            self.ogr_project.get_pr_files_diff(6)
+
+    def test_pr_diff_failing_and_succeding(self):
+        diff = self.ogr_project.get_pr_files_diff(7, retries=1, wait_seconds=0.1)
```
nforro, line 134: Perhaps we could also use flexmock to make sure that the API call is really done twice:

```suggestion
        flexmock(self.ogr_project).should_receive("_call_project_api").twice()
        diff = self.ogr_project.get_pr_files_diff(7, retries=1, wait_seconds=0.1)
```

ogr/services/pagure/pull_request.py
```diff
@@ -191,18 +192,42 @@ def get(project: "ogr_pagure.PagureProject", pr_id: int) -> "PullRequest":
...
+        retries=0,
+        wait_seconds=3,
+    ) -> dict:
+        """Collect all the changes in the PR and deal with ENOPRSTATS exception.
+        While the PR is changing its status from open to merged
+        exception can be raised by the Pagure service.
+        Wait for a while to see if we can manage to access changes adjusting the
+        retries and wait_seconds parameters.
+
+        """
```
nforro, line 207:
```suggestion
        """
        Retrieve pull request diff statistics.
        
        Pagure API tends to return ENOPRSTATS error when a pull request is transitioning
        from open to other states, so you can use `retries` and `wait_seconds` to try to mitigate that.
        
        Args:
            project: Pagure project.
            pr_id: Pull request ID.
            retries: Number of extra attempts.
            wait_seconds: Delay between attempts.
        """
```

tests/integration/pagure/test_pull_requests.py
majamassarini: I tried this but if I mock the project I don't receive anymore the dictionary and the exception... probably this should be checked by requere. If we have two requests, requere should check that they both are used. However I debugged the method and I have seen that both requests are used.

nforro: Ok, let's not comoplicate this.

nforro (approved):

PR #834, opened 2024-01-24
Add getting groups and users with given access rights for Pagure

lbarcziova:
Needed for packit/packit#2088

RELEASE NOTES BEGIN
For Pagure there are 2 new methods available: one for getting users with specified access rights and one for getting members of a group.
RELEASE NOTES END

### DIFF
--- a/ogr/services/pagure/group.py
+++ b/ogr/services/pagure/group.py
@@ -11,7 +11,7 @@ logger = logging.getLogger(__name__)
 class PagureGroup:
     service: "ogr_pagure.PagureService"
 
-    def __init__(self, name: str, raw_group) -> None:
+    def __init__(self, name: str, raw_group: dict) -> None:
         self.name = name
         # see https://pagure.io/api/0/#groups-tab
         self._raw_group = raw_group
--- a/ogr/services/pagure/group.py
+++ b/ogr/services/pagure/group.py
@@ -1,0 +1,24 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import logging
+
+from ogr.services import pagure as ogr_pagure
+
+logger = logging.getLogger(__name__)
+
+
+class PagureGroup:
+    service: "ogr_pagure.PagureService"
+
+    def __init__(self, name: str, raw_group) -> None:
+        self.name = name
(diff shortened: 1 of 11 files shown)


ogr/services/pagure/group.py
```diff
@@ -0,0 +1,24 @@
...
+
+from ogr.services import pagure as ogr_pagure
+
+logger = logging.getLogger(__name__)
+
+
+class PagureGroup:
+    service: "ogr_pagure.PagureService"
+
+    def __init__(self, name: str, raw_group) -> None:
```
nforro, line 14:
```suggestion
    def __init__(self, name: str, raw_group: dict) -> None:
```

ogr/services/pagure/group.py
```diff
@@ -0,0 +1,24 @@
...
+        self.name = name
+        # see https://pagure.io/api/0/#groups-tab
+        self._raw_group = raw_group
```
nforro, line 17: I would rather call this `group_data` or something like that, I find that more self-explanatory.

lbarcziova: we use this terminology in other classes already, that's why I went with this

tests/integration/pagure/test_generic_commands.py
```diff
@@ -119,11 +119,22 @@ def test_pr_permissions(self):
         assert "lachmanfrantisek" in owners
         assert self.ogr_project.can_merge_pr("lachmanfrantisek")
         project = self.service.get_project(
-            repo=f"playground-{self.service.user.get_username()}",
-            namespace=None,
+            repo="license-validate",
+            namespace="copr",
```
nforro, line 123: I'm not sure how good idea it is to depend on a real project we don't have full control over (meaning if something needs to be changed for whatever reason we can't really object against it).

nforro: Even though it wouldn't break the tests thanks to recorded requests, it would be a headache to try to re-record it. But maybe I'm overthinking this.

nforro (approved):

tests/integration/pagure/test_generic_commands.py
lbarcziova: good point, I am going to change it

nforro (approved):

PR #839, opened 2024-02-07
Unify interface for labels

lbarcziova:
Needed for packit/packit-service#2334

RELEASE NOTES BEGIN
Interface for labels was unified and `labels` property for PullRequest and Issue now return list of PRLabel and IssueLabel respectively.
RELEASE NOTES END

### DIFF
--- a/ogr/abstract.py
+++ b/ogr/abstract.py
@@ -332,7 +332,7 @@ class Issue(OgrAbstractClass):
         raise NotImplementedError()
 
     @property
-    def labels(self) -> list:
+    def labels(self) -> list["IssueLabel"]:
         """Labels of the issue."""
         raise NotImplementedError()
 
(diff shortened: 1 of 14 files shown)


ogr/abstract.py
```diff
@@ -2062,3 +2062,37 @@ def get_forks(self) -> Sequence["GitProject"]:
...
+
+
+class IssueLabel(Label):
+    @property
+    def issue(self) -> "Issue":
+        """Issue of issue label."""
+        return self._parent
+
+    def __str__(self) -> str:
+        return "Issue" + super().__str__()
```
mfocko, line 2088: I think that this in combination with the `__str__` in the forge-specific classes will create a rather weird string representation, but in the end… it's still just a string, so I think it's fine.

mfocko (approved):
LGTM

Issue #838, opened 2024-01-31
Resolving paths


mfocko:
### What happened? What is the problem?

GitLab fails to resolve `.` and (potentially) `..` in the paths, this resulted in **not** triggering the TF runs.

Possibly affected methods (as TODO list with confirmed as ticked):
- [x] `GitProject.get_file_content()`
- [ ] `PullRequest.comment()`

### What did you expect to happen?

`ogr` to ignore the `.` and `..` where applicable

### Example URL(s)

Example of sanitization:
https://github.com/packit/packit-service/blob/7337498d73eed935e6196696a518d3ea69ef4ad7/packit_service/worker/helpers/testing_farm.py#L191-L202

This could be probably done in a more generic way in the `packit-service`, that's open for discussion. The linked sanitization is not enough IMO…

### Steps to reproduce

```bash
import os
from ogr import GitlabService

gs = GitlabService(token=os.getenv("GITLAB_TOKEN"))
p = gs.get_project(namespace="efedin", repo="compose-tests")

# generated via f-string in Packit Service
print(p.get_file_content(path="./.fmf/version", ref="84d980d")

# without the `.` prefix
print(p.get_file_content(path=".fmf/version", ref="84d980d")
```

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [X] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

majamassarini:
In ogr we want the same behaviour for github, gitlab and pagure. Fix it in packit if this will change the behaviour for gitlab in ogr.

lbarcziova:
`PullRequest.comment()` is not affected as `filename` there is not event considered for GitLab, closing.

PR #844, opened 2024-02-28
Remove ./ when getting file in GitLab

lbarcziova:
So that the behaviour is consistent among forges. GitLab cannot do this automatically (confirmed that both GitHub and Pagure can).

Related to #838

RELEASE NOTES BEGIN

`GitLabProject.get_file_content` can now correctly handle file paths starting with `./`.
RELEASE NOTES END

### DIFF
--- a/ogr/services/gitlab/project.py
+++ b/ogr/services/gitlab/project.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 import logging
+import os
 from typing import Any, Optional, Union
 
 import gitlab
@@ -361,6 +362,8 @@ class GitlabProject(BaseGitProject):
 
     def get_file_content(self, path, ref=None) -> str:
         ref = ref or self.default_branch
+        # GitLab cannot resolve './'
+        path = os.path.normpath(path)
         try:
             file = self.gitlab_repo.files.get(file_path=path, ref=ref)
             return file.decode().decode()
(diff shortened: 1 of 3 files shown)


tests/integration/gitlab/test_generic_commands.py
```diff
@@ -23,6 +23,16 @@ def test_get_file_content(self):
             "https://github.com/packit-service/ogr\n\ntest1\ntest2\n"
         )
 
+    def test_get_file_content_resolve_dot(self):
+        file = self.project.get_file_content(
+            path="./README.md",
```
mfocko, line 28: I hope that we somehow don't end up with `././` :D

mfocko (approved):

tests/integration/gitlab/test_generic_commands.py
lbarcziova: good point, I changed it to use `normpath`

ogr/services/gitlab/project.py
```diff
@@ -361,6 +362,8 @@ def get_branches(self) -> list[str]:
 
     def get_file_content(self, path, ref=None) -> str:
         ref = ref or self.default_branch
+        # GitLab cannot resolve './'
+        path = os.path.normpath(path)
```
mfocko, line 366: I was looking for this function, but everything I managed to find on the `Path` class was impure /o And this (at least documentation doesn't say a thing) is pure, but under the `os` module /o\ This language is a mess…

mfocko (approved):

lbarcziova:
regate

PR #862, opened 2024-10-01
Fixes related with packit service slowliness

majamassarini:
Should partially fix packit/packit-service#2512


mfocko (approved):
I saw red TF, I got scared we'll have to regenerate the responses :eyes:

lbarcziova (approved):
also thanks for the detailed writedown!

PR #893, opened 2025-03-04
Feat/forgejo/projects

mfocko:
TODO:

- [x] Write new tests or update the old ones to cover new functionality.
- [x] Update doc-strings where appropriate.
- [x] Update or write new documentation in `packit/packit.dev`.
- [x] `ForgejoProject.get_file_content` returns base64-encoded contents
- [x] `ForgejoProject.get_files` needs to be implemented manually as in Pagure…
  - [ ] [forgejo upstream issue](https://codeberg.org/forgejo/forgejo/issues/7328)
  - [ ] [pyforgejo tracking issue](https://codeberg.org/harabat/pyforgejo/issues/11)
  - [ ] [pyforgejo fix](https://codeberg.org/harabat/pyforgejo/pulls/12)
- [x] Improve the partial calls (that reuse `owner, repo`)

Fixes #877

PR #896, opened 2025-03-08
[WIP] Add Forgejo issues, pull_request comments and reactions support

### LINKED ISSUE #880: Implement support for Forgejo issues
- [ ] implement methods for the class inheriting from [`Issue`](https://github.com/packit/ogr/blob/31f2ff50bfa4583ffe108640666f8470c692bef0/ogr/abstract.py#L284)
- [ ] use functionality from https://codeberg.org/harabat/pyforgejo/src/branch/main/pyforgejo/issue/client.py

Client usage docs: https://codeberg.org/harabat/pyforgejo#usage

---
*Part of #866*

mynk8:
TODO:
- [ ] Write new tests or update the old ones to cover new functionality.

Fixes #880 

RELEASE NOTES BEGIN

RELEASE NOTES END


ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+    def author(self) -> str:
+        return self._raw_issue.user.login
+
+    @property
+    def created(self) -> datetime:
+        return self._raw_issue.created_at
+
+    @property
+    def status(self) -> IssueStatus:
+        return IssueStatus[self._raw_issue.state]
```
mfocko, line 59: Are you sure this maps as it should?

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+            body=body,
+            labels=labels,
+            assignees=assignees,
+        )
+        return ForgejoIssue(issue, project)
+
+    @staticmethod
+    def get(project: "forgejo.ForgejoProject", issue_id: int) -> _issue:
+        if not project.has_issues:
+            raise IssueTrackerDisabled
```
mfocko, line 96:
```suggestion
            raise IssueTrackerDisabled()
```

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+
+        try:
+            issue = project.service.api.issue.get_issue(
+                owner=project.namespace, repo=project.repo, index=issue_id,
+            )
+        except Exception as ex:
+            raise OperationNotSupported(f"Issue {issue_id} not found") from ex
```
mfocko, line 103: _operation not supported_ when the issue is not found?

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+    @staticmethod
+    def get_list(
+        project: "forgejo.ForgejoProject",
+        status: IssueStatus = IssueStatus.open,
+        author: Optional[str] = None,
+        assignee: Optional[str] = None,
+        labels: Optional[list[str]] = None,
+    ) -> list["Issue"]:
+        if not project.has_issues:
+            raise IssueTrackerDisabled
```
mfocko, line 115:
```suggestion
            raise IssueTrackerDisabled()
```

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+
+        parameters: dict[str, Union[str, list[str], bool]] = {
+            "state": status if status != IssueStatus.open else "open",
```
mfocko, line 118: I'm not really sure if this produces what you think it does… looking at the definition of `IssueStatus`, it is an `IntEnum`, I think you'll find out when you start testing it.

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+            parameters["created_by"] = author
+        if assignee:
+            parameters["assigned_by"] = assignee
+        if labels:
+            parameters["labels"] = labels
+
+        issues = project.service.api.issue.list_issues(
+            owner=project.namespace, repo=project.repo, **parameters,
+        )
+        return [ForgejoIssue(issue, project) for issue in issues]
```
mfocko, line 131: This part here is a bit tricky… See https://github.com/packit/packit/discussions/2543#discussioncomment-12411705 and related notes…

1. You don't get a full list, cause it is paginated
2. Once you start fetching everything page-by-page, don't construct a list, implement this as an iterator  (I'll be committing similar code for `ForgejoProject` today, so you should be able to see basic outline of how it looks like)

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+        )
+
+    def comment(self, body: str) -> IssueComment:
+        comment = self._issue_api_call(
+            self.project.service.api.issue.create_comment,
+            body=body,
+        )
+        return ForgejoIssueComment(self, comment)
+
+    def _get_all_comments(self) -> list[IssueComment]:
```
mfocko, line 153: Same here with the pagination

ogr/services/forgejo/issue.py
```diff
@@ -1,10 +1,167 @@
...
+        comments = self._issue_api_call(self.project.service.api.issue.get_comments)
+        return [
+            ForgejoIssueComment(raw_comment=comment, parent=self)
+            for comment in comments
+        ]
+
+    def _issue_api_call(self, method, **kwargs):
```
mfocko, line 160: Eh… OK… I would probably even go with something like:

```suggestion
    def _api_call(self, method, **kwargs):
```

mfocko (commented):
IIUC the docs are inherited from the abstract classes, so we're left with just the pagination and tests

ogr/services/forgejo/issue.py
p0tat0chip: I’m still new to this and learning the ropes, so I appreciate your patience. I’ll implement the suggested changes and do my best to improve, though there might still be a few rookie mistakes along the way. Thanks for understanding!

ogr/services/forgejo/issue.py
p0tat0chip: Totally missed it, will do the changes

ogr/services/forgejo/issue.py
p0tat0chip: Did manually test this and had similar doubts, I will check again

mfocko: If you cover this with tests after, it should be fine.

p0tat0chip:
So, I wrote all the tests for Forgejo Issues from the Gitlab ones. Not sure how requre works and test_data need a little guidance there. What's left is to handle pagination

ogr/services/forgejo/issue.py
p0tat0chip: Now using NotFoundError

mynk8:
my bad cleaning this up

mfocko:
Hi, I've pulled in some changes from #893, could you please rebase and see what happens (with the CI)? :)

Also, I've added some utility functions related to the workaround for repeated `owner, repo` in the API, it would be ideal, if you switched to that from your own “hack”:
- pagination (**note**: we have agreed on returning iterables instead of the whole lists, you have implemented it with the lists so far, could you please switch to the iterables? you can also see some of the changes it required in the commits, if you need help, feel free to reach out)
  - function itself
    https://github.com/packit/ogr/blob/9c5d2d5025f47bf34c6f5f8120eaa02173faf3f2/ogr/services/forgejo/utils.py#L7-L14
  - usage
    https://github.com/packit/ogr/blob/9c5d2d5025f47bf34c6f5f8120eaa02173faf3f2/ogr/services/forgejo/project.py#L361-L368
- repeated parameters in the API
  - function
    https://github.com/packit/ogr/blob/9c5d2d5025f47bf34c6f5f8120eaa02173faf3f2/ogr/services/forgejo/project.py#L55-L82

    I think you don't have much choice than to copy the `partial_api` and create another `.api` property on the `ForgejoIssue` that would return a reference to the `issue` API endpoint (rather than the `repository` one as on the project)
  - usage

    https://github.com/packit/ogr/blob/9c5d2d5025f47bf34c6f5f8120eaa02173faf3f2/ogr/services/forgejo/project.py#L109-L111

mynk8:
I will implement them as iterators. Thanks for the suggestions; they were much needed.

mynk8:
a little busy with exams... will take up again in 2 days. I apologise if anything's been blocked due to this.

ogr/services/forgejo/issue.py
```diff
@@ -37,4 +145,46 @@ def get_list(
...
+                project.service.api.issue.list_issues,
+                owner=project.namespace,
+                repo=project.repo,
+                **parameters,
+            )
+            return [ForgejoIssue(issue, project) for issue in issues_paginator]
+        except NotFoundError as ex:
+            if "user does not exist" in str(ex):
+                return []
+            raise OperationNotSupported(f"Failed to list issues {ex}") from ex
```
mfocko, line 172: I don't think it's a good idea, we should wrap everything, so in this case, I would just let any issue get wrapped in the `ForgejoAPIException` rather than specifically handle cases like this.

ogr/services/forgejo/issue.py
```diff
@@ -37,4 +145,46 @@ def get_list(
...
+
+    def close(self) -> "Issue":
+        self.partial_api(self.api.edit_issue)(state="closed")
+        self.__update_info()
```
mfocko, line 176: Same here, you should get an issue from the `edit_issue`

ogr/services/forgejo/issue.py
```diff
@@ -1,18 +1,101 @@
...
+    @property
+    def created(self) -> datetime:
+        return self._raw_issue.created_at
+
+    @property
+    def status(self) -> IssueStatus:
+        return IssueStatus[self._raw_issue.state]
+
+    @property
+    def assignees(self) -> list:
```
mfocko, line 97:
```suggestion
    def assignees(self) -> list[types.PyforgejoUser]:
```

ogr/services/forgejo/issue.py
```diff
@@ -1,18 +1,101 @@
...
+        return getattr(self._raw_issue, "assignees", []) or []
```
mfocko, line 98:
```suggestion
        return self._raw_issue.assignees or []
```

I just checked the pyforgejo, this is way too overcomplicated

ogr/services/forgejo/issue.py
```diff
@@ -37,4 +145,46 @@ def get_list(
...
+        return self
+
+    def add_assignee(self, *assignees: str) -> None:
+        current_assignees = [
+            assignee.login if hasattr(assignee, "login") else assignee
+            for assignee in self.assignees
+        ]
+        updated_assignees = list(set(current_assignees + list(assignees)))
+        self.partial_api(self.api.edit_issue)(assignees=updated_assignees)
+        self.__update_info()
```
mfocko, line 186: `edit_issue` returns updated object

ogr/services/forgejo/issue.py
```diff
@@ -37,4 +145,46 @@ def get_list(
...
+
+    def add_label(self, *labels: str) -> None:
+        self.partial_api(self.api.add_label)(labels=labels, index=self._index)
```
mfocko, line 189: Why do you _hack around_ the index (issue number) above, when you pass it here anyways?

tests/integration/forgejo/base.py
```diff
@@ -0,0 +1,45 @@
...
+                instance_url="https://v10.next.forgejo.org",
+            )
+        return self._service
+
+    @property
+    def project(self):
+        if not self._project:
+            self._project = self.service.get_project(
+                repo="ogr-tests",
+                namespace="manky201",
```
mfocko, line 43: hardcoded username

tests/integration/forgejo/test_issues.py
```diff
@@ -0,0 +1,82 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+from requre.online_replacing import record_requests_for_all_methods
+
+from ogr.abstract import IssueStatus
+from tests.integration.forgejo.base import ForgejoTests
+
+
+@record_requests_for_all_methods()
+class Issues(ForgejoTests):
```
mfocko, line 10: I guess this is mostly for @majamassarini as she started the parametrized tests for forgejo without any classes as we did before, I'm leaning towards keeping that convention rather than falling back to what we had before, but I also have to admit I don't remember how are the latest Forgejo tests done

PR #897, opened 2025-03-10
Build ogr just for stable releases

majamassarini:
in copr packit-stable project

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -33,21 +33,18 @@ jobs:
     trigger: release
     dist_git_branches:
       - fedora-all
-      - epel-9
 
   - job: copr_build
     trigger: pull_request
     targets:
       - fedora-all
       - fedora-all-aarch64
-      - epel-9
     additional_repos:
       - "copr://packit/packit-dev"
   - job: tests
     trigger: pull_request
     targets:
       - fedora-all
-      - epel-9
     tf_extra_params:
       environments:
         - artifacts:
@@ -66,7 +63,6 @@ jobs:
     targets:
       - fedora-all
       - fedora-all-aarch64
-      - epel-9
     project: packit-dev
     list_on_homepage: True
     preserve_project: True
@@ -75,9 +71,8 @@ jobs:
     trigger: commit
     branch: stable
     targets:
-      - fedora-stable
-      - fedora-stable-aarch64
-      - epel-9
+      - fedora-latest-stable
+      - fedora-latest-stable-aarch64
     project: packit-stable
     list_on_homepage: True
     preserve_project: True
@@ -87,7 +82,6 @@ jobs:
     targets:
       - fedora-all
       - fedora-all-aarch64
-      - epel-9
     project: packit-releases
     list_on_homepage: True
     preserve_project: True
@@ -96,7 +90,6 @@ jobs:
     trigger: release
     dist_git_branches:
       - fedora-all
-      - epel-9
 
   # downstream automation:
   - job: koji_build
@@ -108,4 +101,3 @@ jobs:
       - packit
     dist_git_branches:
       - fedora-all
-      - epel-9


.packit.yaml
```diff
@@ -77,7 +77,6 @@ jobs:
     targets:
       - fedora-stable
```
mfocko, line 78: `fedora-stable` still includes F40 that fails

mfocko: Actually not sure what we want to do here… we could probably omit `targets` altogether and deduce from the Copr settings, but that would mean that we need to remove this entry from all of `specfile, ogr, packit`, cause either of them could “reset” the F40 and EL9 back.

majamassarini: I have already started removing things in packit, https://github.com/packit/packit/pull/2549/files. At least we should remove epel9 there. And use `fedora-latest` instead?

nforro: > if we intend to move our deployments with Fedora releases, then I guess we could use `fedora-latest`

Hm, but if we forget will we be able to notice (soon enough) that the images contain some old builds?

mfocko: As long as we're talking about `packit/packit-stable` Copr, I would say it would be enough to deduce the targets from Copr and adjust the release in images and Copr repo at the same time, manually.

mfocko: That also means we should eradicate `targets: […]` in all repos that build in `packit/packit-stable`…

majamassarini: I am not sure I follow, why using `fedora-latest` our images could use old builds?

mfocko: > I am not sure I follow, why using `fedora-latest` our images could use old builds?

If we bump (or not bump) the base image without adjusting the Copr settings (or config here), we will start pulling the dependencies from official Fedora repository (which will not necessarily be the same as us moving the stable and using Copr repo)

lbarcziova: what about going with `fedora-latest-stable`?

lbarcziova (approved):

.packit.yaml
```diff
@@ -108,4 +107,4 @@ jobs:
       - packit
     dist_git_branches:
       - fedora-all
-      - epel-9
+      - epel-10-all
```
nforro, line 110: We need to coordinate this with specfile and packit.

nforro (approved):

.packit.yaml
```diff
@@ -96,7 +95,7 @@ jobs:
     trigger: release
     dist_git_branches:
       - fedora-all
-      - epel-9
+      - epel-10-all
```
nforro, line 98: I would rather do this (same for `propose_downstream`):

```suggestion
      epel-10:
        fast_forward_merge_into:
          - epel-10-branched
```

nforro: But if we are not doing it for Fedora on purpose, then it makes sense to have divergent branches in EPEL 10 too.

majamassarini: Yes, I thought that we could start with the same commits, until we can at least. And than when we are forced we can diverge.

.packit.yaml
nforro: There are no builds because the `epel-10-all` alias doesn't work yet.

majamassarini: I will wait tomorrow, that fedora-distro-aliases should reach stable update, to test this PR properly and also push the packit one.

.packit.yaml
nforro: I mean, without `fast_forward_merge_into`, the `epel10.0` branch will have different commit hashes than `epel10`.

majamassarini:
In the end we decided to remove both epel9 and epel10, since there are missing dependencies ([gitlab bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2343587) and [pygithub bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2351832)).

PR #920, opened 2025-04-13
resolved 56 of the 94 errors

pranav33317:
TODO:

- [] Write new tests or update the old ones to cover new functionality.
- [ ] Update doc-strings where appropriate.
- [ ] Update or write new documentation in `packit/packit.dev`.
- [x] ‹resolved errors occurring by running pre-commit run --all-files›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

Packit now supports automatic ordering of ☕ after all checks pass.

RELEASE NOTES END

### DIFF
--- a/ogr/services/forgejo/project.py
+++ b/ogr/services/forgejo/project.py
@@ -307,6 +307,7 @@ class ForgejoProject(BaseGitProject):
         raise NotImplementedError()
 
     @indirect(ForgejoIssue.get_list)
+    # type: ignore
     def get_issue_list(
         self,
         status: IssueStatus = IssueStatus.open,
@@ -317,10 +318,12 @@ class ForgejoProject(BaseGitProject):
         pass
 
     @indirect(ForgejoIssue.get)
+    # type: ignore
     def get_issue(self, issue_id: int) -> "Issue":
         pass
 
     @indirect(ForgejoIssue.create)
+    # type: ignore
     def create_issue(
         self,
         title: str,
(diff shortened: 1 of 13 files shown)


ogr/services/forgejo/project.py
```diff
@@ -307,6 +307,7 @@ def remove_group(self, group: str) -> None:
         raise NotImplementedError()
 
     @indirect(ForgejoIssue.get_list)
+    # type: ignore
```
gotmax23, line 310:
```suggestion
    # type: ignore[empty-body]
```

ogr/services/pagure/release.py
```diff
@@ -35,7 +35,7 @@ def url(self) -> Optional[str]:
 
     @property
     def created_at(self) -> datetime.datetime:
-        return None
+        return datetime.datetime.now
```
gotmax23, line 38: It looks like you forgot parentheses. This returns a function object, not a datetime. In any case, I'm not sure returning the current time is right here. Probably better to change the type signature to `datetime.datetime | None` here and in the base class to reflect that some implementations don't provide this information.

ogr/services/forgejo/project.py
gotmax23: Same applies elsewhere. It's better to pass an explicit error code when to `type: ignore` pragmas, but maybe it doesn't matter so much here.

ogr/services/pagure/pull_request.py
```diff
@@ -26,7 +26,9 @@
 
 class PagurePullRequest(BasePullRequest):
     _target_project: "ogr_pagure.PagureProject"
-    _source_project: "ogr_pagure.PagureProject" = None
+    if ogr_pagure.PagureProject is None:
+        _source_project = ""
+    _source_project: "ogr_pagure.PagureProject"
```
gotmax23, line 31:
```suggestion
    _source_project: Optional[ogr_pagure.PagureProject]
```

PR #937, opened 2025-08-06
docs: extend the tmt documentation based on #936

mfocko:
TODO:
- [x] Mention the possibility of using an environment file.

### DIFF
--- a/plans/README.md
+++ b/plans/README.md
@@ -89,9 +89,9 @@ Explanation of the `tmt` command:
 > file.
 
 Afterwards the generated responses can be found in the test directory
-(`???` varies):
+(`<ID>` varies):
 
-    /var/tmp/tmt/run-???/plans/full/discover/default-0/tests/
+    /var/tmp/tmt/<ID>/plans/full/discover/default-0/tests/
 
 > [!WARNING]
 >
@@ -99,7 +99,7 @@ Afterwards the generated responses can be found in the test directory
 > the `tmt clean` command. You can either:
 >
 > - clean the last run with `tmt clean -l`, or
-> - specify a concrete ID of the run with `tmt clean --id ID`.
+> - specify a concrete ID of the run with `tmt clean --id <ID>`.
 
 ## Additional resources
 


plans/README.md
```diff
@@ -1,12 +1,80 @@
...
+(`???` varies):
+
+    /var/tmp/tmt/run-???/plans/full/discover/default-0/tests/
+
+> [!WARNING]
+> Don't forget to clean the directories afterwards. This can be done by running
+> the `tmt clean` command. You can either:
+>
+> - clean the last run with `tmt clean -l`, or
+> - specify a concrete ID of the run with `tmt clean --id ID`.
```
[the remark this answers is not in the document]
mfocko: it should be though 👀 isn't it? /o\ I thought only the number is ID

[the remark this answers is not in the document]
betulependule: I thought so too, but when I used only the number `tmt clean --id 024`, then I got the following:
```
clean
guests

Directory '/var/tmp/tmt/024' does not exist.
```

However, `run-024` definitely exists, so I tried `tmt clean --id run-024` and that seemed to work:
```
clean
guests
        finish
        
            summary: 0 tasks completed
            container: stopped
            container: removed
            container: network removed
runs
images
    testcloud
        warn: Directory '/var/tmp/tmt/testcloud/images' does not exist.
```

betulependule:
Aside from the small note regarding IDs of tests runs, everything looks good to me.

.env.template
```diff
@@ -0,0 +1,11 @@
+# Defaults to all tests
+TEST_TARGET: "."
```
betulependule, line 2: Are you sure this is what the syntax should be? When I ran `tmt -vv run -a -k -e @.env provision -h container plan -n full report -h display` mentioned below, I got an error:

> 
> /var/tmp/tmt/run-010
> 
> Failed to extract variables from 'dotenv' format.
> 
> The exception was caused by 1 earlier exceptions
> 
> Cause number 1:
> 
>     not enough values to unpack (expected 2, got 1)
> 

However, I tried to edit the syntax inside the .env file from `TEST_TARGET: "."` to `TEST_TARGET="."` (without spaces before and after `=`, those cause the same error as well) and then the tests ran without a hitch. I had to change the syntax for all of the variables inside the .env file.

mfocko: And I have an exact opposite problem 🤦 it doesn't work for me with the dotenv format /o\ jfc

mfocko: Just out of interest, when you do `tmt --version` what do you get? I have `1.51.0`

mfocko: should be fixed now

betulependule (approved):

Issue #934, opened 2025-07-17
Cannot build test image using ansible-bender


betulependule:
### What happened? What is the problem?

Trying to build a test image using `make build-test-image` (as described in CONTRIBUTING.md) in order to run tests results in the following error:

```
There was an error during execution: no python interpreter was found in the base image "fedora:latest", you can specify the path via CLI option --python-interpreter
make: *** [Makefile:9: build-test-image] Error 2
```

Makefile should be edited so that `make build-test-image` and subsequent `make check-in-container` work. As far as I can tell, there is no public image of Fedora containing a Python interpreter available. As suggested by nforro, an option is to create a local image of Fedora using podman and subsequently install a Python interpreter, but it would be a lazy solution.

A cleaner solution would be to avoid needing to use ansible-bender and create a local image with ansible-playbook (use packit/base:fedora) similarly to how's it's done in [packit](https://github.com/packit/packit/blob/main/Containerfile.tests).

It's possible to run tests as described in plans/README.md using tmt for the time being.

### Steps to reproduce

```bash
1. make build-test-image
```

### Workaround

- [x] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

PR #941, opened 2025-08-26
Fix local image build for testing

betulependule:
1. Due to previous issues with using `fedora:latest` image that lacked a Python interpreter necessary to run tests, the test image is now created using `Containerfile.tests`. All necessary setup is done inside this image using recipe.yaml's content, which has been split into multiple files.

2. The Makefile rule `check-pypi-packaging` is now fixed as the old rule relied on a now non-existent file `setup.py`. The necessary Python module `build` was added to requirements installed inside test image.

3. `Httpx` is now used when making requests from our Forgejo implementation (for the
sake of consistency with the pyforgejo that uses `httpx` too). Forgejo tests were not recorded previously when using `requests`.

Fixes [#934](https://github.com/packit/ogr/issues/934)


mfocko:
I've removed some of the unnecessary files, also reading the description now… I'm for removing the `recipe.yaml`, this is the only repo that used `ansible-bender` for building the container images (and tbf, I don't even see it packaged in nixpkg, so it doesn't work for me anyways; would have to install it via pip(x) which I'd like to avoid…)

Given the “complexity” of the `ogr` project, this “machinery” still feels quite heavy, but having a quick look at it, I'm not really sure whether it's worth burning the time on refining the CI that seems to be working…

In your commit I've also noticed some references to Zuul, i.e., some of the tasks are shared between local and Zuul CI, so, IMO it is definitely possible to merge the `files/tasks` and `files/test-tasks`, on one side I don't like the fragmentation, but on the other it feels like some stuff is going to break, so I'm not really convinced it's worth it.

betulependule:
> I've removed some of the unnecessary files, also reading the description now… I'm for removing the `recipe.yaml`, this is the only repo that used `ansible-bender` for building the container images (and tbf, I don't even see it packaged in nixpkg, so it doesn't work for me anyways; would have to install it via pip(x) which I'd like to avoid…)

Great job, thanks.

> Given the “complexity” of the `ogr` project, this “machinery” still feels quite heavy, but having a quick look at it, I'm not really sure whether it's worth burning the time on refining the CI that seems to be working…

I'm not sure what other way to do this. I'm very much new to ContainerFile and such. :sweat_smile: 

> In your commit I've also noticed some references to Zuul, i.e., some of the tasks are shared between local and Zuul CI, so, IMO it is definitely possible to merge the `files/tasks` and `files/test-tasks`, on one side I don't like the fragmentation, but on the other it feels like some stuff is going to break, so I'm not really convinced it's worth it.

At the very least, it would look nicer to have these folders merged together. As far as I can tell, there are a couple of files with same file names and similar code inside, so these would need to be edited. :eyes:

mfocko (approved):
Ideally remove the `recipe.yaml`, since it's not used.

PR #981, opened 2026-02-26
Count requests per namespace

### LINKED ISSUE #2912: Create a metric for counting Packit requests per project
So that we can have a better understanding of the rate-limit issues we have.

We want to track the requests we do per project.
The idea is described [here](https://github.com/packit/research/blob/2400cbbae84db38d3840ba4d1d274bc88095be1e/research/monitoring/github-api-call-inventory/index.md?plain=1#L4) (approach 1). It should be applied to all the services not only Github.

To implement it we can, for example, inject the Prometeus objects into the Ogr services classes and increase a counter for every request in ogr.

majamassarini:
This implementation isn't thread safe. I don't think we need to risk a deadlock just for tracking the trend of requests in a namespace. We can lose some of them.
Also, the aggregation of counts across different workers needs to be done in Grafana.

Should fix https://github.com/packit/packit-service/issues/2912
Merge before https://github.com/packit/packit-service/pull/3026

### DIFF
--- a/ogr/metrics.py
+++ b/ogr/metrics.py
@@ -1,0 +1,132 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+"""
+Metrics tracking for OGR API calls.
+
+This module provides a lightweight metrics tracking system to count
+API calls per service type and namespace. Metrics are stored in-memory
+and are NOT thread-safe or process-synchronized.
+
+The track_ogr_request decorator can be applied to GitProject methods
+to automatically track API calls.
+"""
+
+import functools
+import logging
+from collections import defaultdict
+from typing import Callable, TypeVar
+
+logger = logging.getLogger(__name__)
+
+F = TypeVar("F", bound=Callable)
+
+
+class RequestMetricsTracker:
+    """
+    Tracker for counting API calls per namespace and service type.
+
+    Note: This class is NOT thread-safe and does NOT synchronize across processes.
+    In a multi-worker environment, each worker maintains its own metrics, and some
+    request counts may be lost. This is acceptable for approximate metrics tracking.
+    The aggregation of data from multiple workers can be done in grafana dashboard.
+    """
+
+    def __init__(self):
+        """Initialize the metrics tracker."""
+        self._counts: dict[tuple[str, str], int] = defaultdict(int)
+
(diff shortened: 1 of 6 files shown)


betulependule (commented):
I like this implementation. 

Regarding it not being thread-safe, I'm thinking, can't we just add locks whenever the `_counts` variable is being written to and read from? It doesn't seem like much of an issue, but I might be missing something. I understand that it would be nice to avoid introducing a deadlock to the code. 😅

mfocko:
Why are we not going with metaclasses here? it might be easier to just wrap everything, like we do for the exceptions, instead of this copy-paste…

At the same time, the simple tracking of `github` / `gitlab` seems to be enough, but with both GitLab and Forgejo there can be self-hosted instances, which kinda breaks the counters?

I'm also a bit afraid that this is something that's quite obviously service-specific, so I'm not sure how it fits the ogr in general.

majamassarini:
> Why are we not going with metaclasses here? it might be easier to just wrap everything, like we do for the exceptions, instead of this copy-paste…

It would be nice to be able to use metaclasses, but I don't think we have a reliable/simple way to decide which methods need wrapping using metaclasses. This way is clearer/more explicit, from my point of view.

> At the same time, the simple tracking of `github` / `gitlab` seems to be enough, but with both GitLab and Forgejo there can be self-hosted instances, which kinda breaks the counters?

True, I missed that. Thanks! I can add the project url.
 
> I'm also a bit afraid that this is something that's quite obviously service-specific, so I'm not sure how it fits the ogr in general.

I partly agree. If the service weren't using only ogr, we'd need more logic there. But since most requests (if not all) go through ogr, I don't see why not. Counting requests where they're made seems correct to me. Also, ogr is a library, so this feature could be useful to other services.

majamassarini:
> Regarding it not being thread-safe, I'm thinking, can't we just add locks whenever the `_counts` variable is being written to and read from? It doesn't seem like much of an issue, but I might be missing something. I understand that it would be nice to avoid introducing a deadlock to the code. 😅

I think the implementation would be simple enough, but I'm concerned we might encounter deadlocks occasionally; when a task fails, a worker is killed, or in other edge cases. I'd like to avoid these situations since we only need a general trend of our request volume.

Issue #984, opened 2026-04-29
Add support for checking permissions beyond merge (e.g. triage)


sergio-correia:
### Description

`can_merge_pr(username)` checks if a user has `admin` or `write` permission on a GitHub repository. `packit-service` uses this to decide who can trigger commands like `/packit test` on PRs, but `"can merge"` is not the right question for `"can interact with the bot"`.

GitHub's `"triage"` role grants permissions that are conceptually sufficient for triggering CI (close/reopen issues and PRs, request reviews, manage labels) without granting write access to repository contents.

### Benefit

This would allow project maintainers to give collaborators the `triage` role instead of `write`, letting them trigger `/packit test` and similar commands without granting unnecessarily broad repository access.

### Importance

Today, the only way to let someone trigger packit commands is to give them `write` access, which violates least-privilege. There is no way to distinguish `"can trigger CI"` from `"can push code"` in the permission model.

Proposal:

Add a method that checks against a broader set of permissions. Either a targeted one:

```python
def can_trigger_ci(self, username: str) -> bool:
    # accepts {"admin", "write", "maintain", "triage"}
```

or a more general one using GitHub's permission hierarchy (`read` < `triage` < `write` < `maintain` < `admin`):

```python
def has_at_least_permission(self, username: str, level: str) -> bool:
```

This would let consumers like `packit-service` use the appropriate check for the action being authorized.

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [x] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

majamassarini:
Thanks for reaching out.
I believe you have a point here and since you checked "I am willing to submit a pull request for this issue", would you like to submit a PR for it and test it with your projects using our staging instance? We would be happy to assist you.

sergio-correia:
Thanks, @majamassarini. I submitted #985

PR #985, opened 2026-04-30
ogr: add has_permission() for access-level checks

sergio-correia:
packit-service currently uses can_merge_pr() to decide who may trigger CI commands like /packit test. This requires write or admin access, but GitHub's triage role — which grants issue/PR management without code push rights — is a better fit for the CI trigger use case. There is no ogr method to check "does this user have at least triage-level access?"

Add has_permission(username, access_level) to GitProject, implemented for all four forges. Each backend uses its own native permission hierarchy for the "at least" comparison, avoiding the IntEnum ordering mismatch between AccessLevel (maintain=5 > admin=4) and GitHub's actual hierarchy (admin > maintain). Forges that lack a triage equivalent (Pagure, Forgejo) conservatively fall back to the next higher level rather than degrading to read/ticket, which would bypass the gate on public repositories.

TODO:

- [x] Write new tests or update the old ones to cover new functionality.
- [x] Update doc-strings where appropriate.
- [x] Update or write new documentation in `packit/packit.dev`.

Each forge uses its own native permission hierarchy for the "at least"
comparison rather than relying on the AccessLevel IntEnum ordering,
which doesn't match GitHub's hierarchy (admin > maintain, but
AccessLevel.maintain=5 > AccessLevel.admin=4).

Forges without a triage equivalent (Pagure, Forgejo) conservatively
fall back to write/commit level — not read/ticket — to avoid bypassing
the permission gate on public repos. This means
_ACCESS_LEVEL_TO_*_PERM intentionally diverges from the existing
access_dict for triage on those forges.

RELEASE NOTES BEGIN

ogr now provides has_permission(username, access_level) to check
whether a user has at least a given access level on a repository,
enabling finer-grained permission checks (e.g., triage) beyond
the existing can_merge_pr().

RELEASE NOTES END

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -63,6 +63,7 @@ In case you find any error, please [create a new issue](https://github.com/packi
 | `which_groups_can_merge_pr`   |   ✘    |   ✘    |            ✔            |    ✘    |
 | `get_pr_files_diff`           |   ✘    |   ✘    |            ✔            |    ✘    |
 | `get_users_with_given_access` |   ✘    |   ✘    |            ✔            |    ✔    |
+| `has_permission`              |   ✔    |   ✔    |            ✔            |    ✔    |
 
 ## User
 
--- a/ogr/abstract/access_level.py
+++ b/ogr/abstract/access_level.py
@@ -8,13 +8,17 @@ class AccessLevel(IntEnum):
     """
     Enumeration representing an access level to the repository.
 
-    | Value from enumeration | GitHub   | GitLab                  | Pagure |
-    | ---------------------- | -------- | ----------------------- | ------ |
-    | `AccessLevel.pull`     | pull     | guest                   | ticket |
-    | `AccessLevel.triage`   | triage   | reporter                | ticket |
-    | `AccessLevel.push`     | push     | developer               | commit |
-    | `AccessLevel.admin`    | admin    | maintainer              | commit |
-    | `AccessLevel.maintain` | maintain | owner (only for groups) | admin  |
+    | Value from enumeration | GitHub   | GitLab                  | Pagure | Forgejo |
+    | ---------------------- | -------- | ----------------------- | ------ | ------- |
+    | `AccessLevel.pull`     | pull     | guest                   | ticket | read    |
+    | `AccessLevel.triage`   | triage   | reporter                | ticket | read    |
+    | `AccessLevel.push`     | push     | developer               | commit | write   |
+    | `AccessLevel.admin`    | admin    | maintainer              | commit | admin   |
+    | `AccessLevel.maintain` | maintain | owner (only for groups) | admin  | owner   |
+
+    Note: ``has_permission()`` uses conservative fallbacks for forges
+    that lack a triage equivalent — a triage check requires commit
+    on Pagure and write on Forgejo, not ticket/read.
     """
 
     pull = 1
--- a/ogr/services/forgejo/project.py
+++ b/ogr/services/forgejo/project.py
@@ -49,6 +49,25 @@ class ForgejoProject(BaseGitProject):
         None: "",
     }
 
+    _FORGEJO_PERM_RANK: ClassVar[dict[str, int]] = {
+        "none": -1,
+        "read": 0,
+        "write": 1,
+        "admin": 2,
+        "owner": 3,
+    }
+
+    # Intentionally differs from access_dict: triage maps to "write"
+    # (not "read") because "read" would bypass the permission gate on
+    # public repos.
+    _ACCESS_LEVEL_TO_FORGEJO_PERM: ClassVar[dict[AccessLevel, str]] = {
+        AccessLevel.pull: "read",
+        AccessLevel.triage: "write",
+        AccessLevel.push: "write",
+        AccessLevel.admin: "admin",
+        AccessLevel.maintain: "owner",
+    }
+
     def __init__(
         self,
         repo: str,
@@ -287,6 +306,28 @@ class ForgejoProject(BaseGitProject):
             repo=self.repo,
             collaborator=username,
         ).permission in ("owner", "admin", "write")
+
+    def has_permission(self, username: str, access_level: AccessLevel) -> bool:
+        try:
+            perm = self.api.repo_get_repo_permissions(
+                owner=self.namespace,
+                repo=self.repo,
+                collaborator=username,
+            ).permission
+        except NotFoundError:
+            return False
+        perm = perm.lower() if perm else "none"
+        if perm not in self._FORGEJO_PERM_RANK:
+            logger.warning(
+                "Unknown Forgejo permission %r for user %s on %s/%s; denying",
+                perm,
+                username,
+                self.namespace,
+                self.repo,
+            )
+            return False
+        required = self._ACCESS_LEVEL_TO_FORGEJO_PERM[access_level]
+        return self._FORGEJO_PERM_RANK[perm] >= self._FORGEJO_PERM_RANK[required]
 
     def get_users_with_given_access(self, access_levels: list[AccessLevel]) -> set[str]:
         access_levels_forgejo = [
--- a/ogr/services/github/project.py
+++ b/ogr/services/github/project.py
@@ -49,6 +49,23 @@ class GithubProject(BaseGitProject):
     # Permission levels that can merge PRs
     CAN_MERGE_PERMS: ClassVar[set[str]] = {"admin", "write"}
 
+    _GITHUB_PERM_RANK: ClassVar[dict[str, int]] = {
+        "none": -1,
+        "read": 0,
+        "triage": 1,
+        "write": 2,
+        "maintain": 3,
+        "admin": 4,
+    }
+
+    _ACCESS_LEVEL_TO_GITHUB_PERM: ClassVar[dict[AccessLevel, str]] = {
+        AccessLevel.pull: "read",
+        AccessLevel.triage: "triage",
+        AccessLevel.push: "write",
+        AccessLevel.admin: "admin",
+        AccessLevel.maintain: "maintain",
+    }
+
     def __init__(
         self,
         repo: str,
@@ -250,6 +267,24 @@ class GithubProject(BaseGitProject):
             self.github_repo.get_collaborator_permission(username)
             in self.CAN_MERGE_PERMS
         )
+
+    def has_permission(self, username: str, access_level: AccessLevel) -> bool:
+        try:
+            perm = self.github_repo.get_collaborator_permission(username)
+        except UnknownObjectException:
+            return False
+        perm = perm.lower() if perm else "none"
+        if perm not in self._GITHUB_PERM_RANK:
+            logger.warning(
+                "Unknown GitHub permission %r for user %s on %s/%s; denying",
+                perm,
+                username,
+                self.namespace,
+                self.repo,
+            )
+            return False
+        required = self._ACCESS_LEVEL_TO_GITHUB_PERM[access_level]
+        return self._GITHUB_PERM_RANK[perm] >= self._GITHUB_PERM_RANK[required]
 
     def _get_collaborators_with_permission(self) -> dict:
         """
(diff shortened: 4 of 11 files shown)


ogr/services/gitlab/project.py
```diff
@@ -193,6 +201,26 @@ def who_can_merge_pr(self) -> set[str]:
...
+        required_access = self._ACCESS_LEVEL_TO_GITLAB[access_level]
+        try:
+            users = self.service.gitlab_instance.users.list(username=username)
+            user = next(
+                (u for u in users if u.username.lower() == username.lower()),
+                None,
+            )
+            if user is None:
+                return False
+            member = self.gitlab_repo.members_all.get(user.id)
```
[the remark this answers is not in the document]
sergio-correia: Good catch. Updated to use `getattr` with `members` as fallback.

ogr/services/pagure/project.py
```diff
@@ -255,6 +274,26 @@ def can_merge_pr(self, username) -> bool:
...
+        ]
+        try:
+            # Check direct user access first to avoid O(G) group expansion
+            project = self.get_project_info()
+            access_users = project.get("access_users", {})
+            for perm in qualifying:
+                if username in access_users.get(perm, []):
+                    return True
+            # Fall back to group expansion
+            return username in self._get_users_with_given_access(qualifying)
```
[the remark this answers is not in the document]
sergio-correia: Agreed. Refactored to reuse the already-fetched project info and iterate groups directly with early exit.

ogr/services/gitlab/project.py
```diff
@@ -193,6 +201,31 @@ def who_can_merge_pr(self) -> set[str]:
...
+                self.gitlab_repo.members,
+            )
+            member = member_manager.get(user.id)
+            member_access = (
+                member["access_level"]
+                if isinstance(member, dict)
+                else member.access_level
+            )
+            return member_access >= required_access
+        except gitlab.exceptions.GitlabOperationError:
```
majamassarini:
```suggestion
        except (gitlab.exceptions.GitlabGetError, gitlab.exceptions.GitlabOperationError):
```
`GitlabGetError` may not inherit from `GitlabOperationError` in rawhide, let's see if this fix the failing tests

majamassarini (changes requested):
Thanks a lot for your contribution!

I have only one small change proposal that will hopefully fix the failing rawhide tests.

This is a nicely done first step toward addressing issue #984. However this alone would not fix it. A change in `packit-service` code is needed for that. Could you please remove "Fixes #984", from the PR description 🙏🏻.

ogr/services/gitlab/project.py
sergio-correia: I followed your suggestion here.

sergio-correia:
Thanks for the review. I updated the PR as per the suggestion, and also updated the description to remove the `Fixes:` clause. I see there are still rawhide issues, but I am not sure they are related to this change.

ogr/abstract/access_level.py
```diff
@@ -8,13 +8,17 @@ class AccessLevel(IntEnum):
...
+    | ---------------------- | -------- | ----------------------- | ------ | ------- |
+    | `AccessLevel.pull`     | pull     | guest                   | ticket | read    |
+    | `AccessLevel.triage`   | triage   | reporter                | ticket | read    |
+    | `AccessLevel.push`     | push     | developer               | commit | write   |
+    | `AccessLevel.admin`    | admin    | maintainer              | commit | admin   |
+    | `AccessLevel.maintain` | maintain | owner (only for groups) | admin  | owner   |
+
+    Note: ``has_permission()`` uses conservative fallbacks for forges
+    that lack a triage equivalent — triage maps to push/write level
+    on Pagure and Forgejo, not to ticket/read.
```
nforro: I'm confused about this comment. Why does the table list `ticket/read` then? Also, Shouldn't it be `commit/write` rather than `push/write`?

nforro:
> I see there are still rawhide issues, but I am not sure they are related to this change.

I believe they are related. You most likely need to regenerate some test data files, as the error message tells you to - by removing the outdated files, running tests locally and committing the newly generated ones.

ogr/abstract/access_level.py
sergio-correia: Good catch, fixed. The note now says "requires commit on Pagure and write on Forgejo" instead of the confusing "push/write". The table shows the assignment mapping. The note explains where has_permission() diverges.

sergio-correia:
> > I see there are still rawhide issues, but I am not sure they are related to this change.
> 
> I believe they are related. You most likely need to regenerate some test data files, as the error message tells you to - by removing the outdated files, running tests locally and committing the newly generated ones.

The Forgejo instance (v10.next.forgejo.org) is down, returning 404 on all endpoints. Same failure on main, as far as I can see. Can't re-record until it's back.

majamassarini:
> I believe they are related. You most likely need to regenerate some test data files, as the error message tells you to - by removing the outdated files, running tests locally and committing the newly generated ones.

If test data must be re-generated, I don't understand why only for rawhide and no for other distros. I will try to investigate more on this.

majamassarini:
Forgejo test failing on fedora rawhide should be fixed by: https://github.com/packit/ogr/pull/986

ogr/services/forgejo/project.py
```diff
@@ -288,6 +307,28 @@ def can_merge_pr(self, username: str) -> bool:
             collaborator=username,
         ).permission in ("owner", "admin", "write")
 
+    def has_permission(self, username: str, access_level: AccessLevel) -> bool:
```
mfocko, line 310: This will very likely be an issue. Trying it out locally I have managed to get the following error message out of it:
```
status_code: 403
message: >
  Only admins can query all permissions, repo admins can query all repo
  permissions, collaborators can query only their own
```

Unfortunately the related issue (to this PR) of service is not linked, but in my perspective this will be a big blocker for supporting Forgejo as an upstream forge in Packit Service, as this also goes against our _least privilege_ “rule” for repos where we provide CI, e.g., on dist-git we **do not** possess **any** privileges.

ogr/services/github/project.py
```diff
@@ -251,6 +268,24 @@ def can_merge_pr(self, username) -> bool:
             in self.CAN_MERGE_PERMS
         )
 
+    def has_permission(self, username: str, access_level: AccessLevel) -> bool:
```
mfocko, line 271: Also, based on the local testing, similar issue as with the Forgejo.

Packit Service is OK here, since it is a GitHub App. However for general usage, with a _classic PAT_ (personal access token) this endpoint yields `403`.

The `403` status is even bigger issue, because of the utterly poor design choices done by GitHub to return `403` even for _rate-limiting_ which classifies **this** _Forbidden_ status code as _to be retried_ in ogr, basically resulting in an infinite loop…

This should be definitely reflected in docs, if not checked manually and handled with a proper exception (IIUC only repo admins, same as with Forgejo, are allowed to query this endpoint).

ogr/services/gitlab/project.py
```diff
@@ -193,6 +201,34 @@ def who_can_merge_pr(self) -> set[str]:
...
         return username in self.who_can_merge_pr()
 
+    def has_permission(self, username: str, access_level: AccessLevel) -> bool:
+        required_access = self._ACCESS_LEVEL_TO_GITLAB[access_level]
+        try:
+            users = self.service.gitlab_instance.users.list(username=username)
+            user = next(
+                (u for u in users if u.username.lower() == username.lower()),
+                None,
+            )
```
mfocko, line 211:
```suggestion
            users = self.service.gitlab_instance.users.list(username=username, iterator=True)
            user = next(users, None)
```

Also, just out of interest, why is there conversion from a `list` to a generator including a filtering? I would've expected the `.list(username=username)` to handle the filtering?

ogr/services/gitlab/project.py
```diff
@@ -193,6 +201,34 @@ def who_can_merge_pr(self) -> set[str]:
...
+                return False
+            member_manager = getattr(
+                self.gitlab_repo,
+                "members_all",
+                self.gitlab_repo.members,
+            )
+            member = member_manager.get(user.id)
+            member_access = (
+                member["access_level"]
+                if isinstance(member, dict)
```
mfocko, line 222: How can this happen? 👀

COMPATIBILITY.md
```diff
@@ -63,6 +63,7 @@ In case you find any error, please [create a new issue](https://github.com/packi
 | `which_groups_can_merge_pr`   |   ✘    |   ✘    |            ✔            |    ✘    |
 | `get_pr_files_diff`           |   ✘    |   ✘    |            ✔            |    ✘    |
 | `get_users_with_given_access` |   ✘    |   ✘    |            ✔            |    ✔    |
+| `has_permission`              |   ✔    |   ✔    |            ✔            |    ✔    |
```
mfocko, line 66: No need to add this, the table is used only when there are features that are missing in some forges, i.e., at least one forge doesn't support the feature.

ogr/services/pagure/project.py
```diff
@@ -47,6 +47,25 @@ class PagureProject(BaseGitProject):
...
+    # has_permission uses "at least" threshold semantics rather than
+    # the assignment semantics of access_dict.
+    _ACCESS_LEVEL_TO_PAGURE_PERM: ClassVar[dict[AccessLevel, str]] = {
+        AccessLevel.pull: "ticket",
+        AccessLevel.triage: "commit",
+        AccessLevel.push: "commit",
+        AccessLevel.admin: "admin",
+        AccessLevel.maintain: "admin",
+    }
+
```
mfocko, line 68: IMO it would probably make sense to make a Pagure-only dataclass… oh i see… ew… never mind

mfocko (commented):
Hi, sorry for the late response.

We've discussed this PR with @majamassarini off GitHub, since there were some issue that I anticipated coming up :/

Overall I kind of understand the point of **not** exposing the maintainers that have higher privileges, at least from the security perspective.

I have tested locally what you can and cannot do with your implementation:
* GitHub PAT (personal access token) — you can check “permissions”¹ of repos that you own (potentially admin, I did not inspect it with such granularity), otherwise you get `403`²
* GitHub App — generally seems to be working, yields 403 for suspended repos and a separate exception for non-enabled repos (where the App has not been installed)
* GitLab — seems OK
* Forgejo — allows admins (of instance; based on my understanding) to query all permissions, repo admins to query on repo and finally contributors to query **only their own**
* Pagure — not investigated; IIRC the use-case for this are permission checks on the **upstream** forges, we do not support Pagure as an upstream forge

The limitations should be clearly documented IMO, especially since they are different for each forge and also have their own nuances, e.g., PAT vs GitHub App.

As for the testing, I'm, as always, leaning towards consistent tests here, therefore integration tests and trying to avoid mocking as much as possible. OTOH in this case, I'm afraid, it doesn't really help anyways, because we generally test on repos that we own (easier to perform actions, no PITA with the privileges), but in this case… it kinda hides the issues I have listed above.

It would be for the best to have integration tests, so that we are sure the functionality works as intended at least. I am aware the testing experience with _requre_ ain't the best, so after we're OK with the implementation/docs, we can possibly take it over to speed it up and alleviate the pain on your end :)

Overall, apart from tests and the design issues of the forges, LGTM :) Thanks a lot for your contribution.

¹ NB with GitHub the permissions are quite complicated… GitHub supports (original) branch protections which are quite complex (you can allow specific people, specific CI checks, bypassing, etc.) and additionally they have moved on to “rulesets” which work in a similar way. In general I doubt it's possible to replicate the behaviour that GitHub allows (I don't think it's safe (for GitHub) to expose branch protections / rulesets outside of the repo owners)

² That `403` is quite annoying, because of another GitHub's decision to use `403` **also** for rate-limiting which we specifically try to handle. I'm not even going to start the rant on the fact that during my testing when I got `403` I also got the rate-limit header to suggest a retry after 50 minutes or so. I am not sure what the “engineers” at GitHub are smoking, but it's pretty good stuff. (And it also justifies projects moving away from GitHub…)

PR #986, opened 2026-05-20
Add retry responses for pyforgejo 2.0.7 in forgejo test cassette

majamassarini:
pyforgejo changed its default retry behavior between versions 2.0.4 and 2.0.7:

- pyforgejo 2.0.4: max_retries defaults to 0 (no retries, 1 HTTP call)
- pyforgejo 2.0.7: max_retries defaults to 2 (2 retries, 3 HTTP calls total)

The test_issue_no_such_assignee test expects a 500 Internal Server Error response when attempting to add a non-existent user as an assignee. This is the actual behavior of the Forgejo API, and the test verifies that ogr properly converts this to a GitForgeInternalError.

Compatibility with both versions:
- pyforgejo 2.0.4 (F43): Uses first response, leaves 2 unused (safe)
- pyforgejo 2.0.7 (Rawhide): Consumes all 3 responses during retries

Fixes test failure in Fedora Rawhide with python3-pyforgejo-2.0.7.

### DIFF
--- a/tests/integration/forgejo/test_data/test_issues/Issues.test_issue_no_such_assignee.yaml
+++ b/tests/integration/forgejo/test_data/test_issues/Issues.test_issue_no_such_assignee.yaml
@@ -303,3 +303,69 @@ httpx._client:
             x-frame-options: SAMEORIGIN
           next_request: null
           status_code: 500
+      - metadata:
+          latency: 0.2674534320831299
+          module_call_list:
+          - unittest.case
+          - requre.record_and_replace
+          - tests.integration.forgejo.test_issues
+          - ogr.abstract.exception
+          - ogr.services.forgejo.issue
+          - pyforgejo.issue.client
+          - pyforgejo.issue.raw_client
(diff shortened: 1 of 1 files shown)


mfocko:
It returns `500` for non-existing user?

mfocko (approved):
In general I don't have any issue with merging this, but I'm also going to have a look at the Forgejo itself. Based on the API docs, I would suspect this should yield `412 — APIError is error format response`.

majamassarini:
> It returns `500` for non-existing user?

yes, you can see it in the registered file ( I mean in those lines I haven't touched)

majamassarini:
> In general I don't have any issue with merging this, but I'm also going to have a look at the Forgejo itself. Based on the API docs, I would suspect this should yield `412 — APIError is error format response`.

Ok thanks! I will merge this one as it is so that the rawhide tests start working again.
If the behaviour changes  in a future version of forgejo we will know where to change the status code :)

PR #988, opened 2026-07-15
chore(pyproject): bump python versions

mfocko:
Trying to run it locally I get yelled at because pyforgejo doesn't support Python 3.9.

Looking at the current Python in CentOS 10 (where we still release updates into EPEL10), the version is 3.12.

As for the current Fedora, it's 3.14, therefore add all these versions to classifiers and bump the minimal requirement from 3.9 to 3.12 as right now it is the minimal version we actively maintain.


betulependule (approved):
LGTM

PR #991, opened 2026-07-17
Fix a bug in relation to the Forgejo "error" commit flag

betulependule:
The internal `_states` dictionary contained improper mapping of the "canceled" state to `CommitStatus.error`. Forgejo doesn't support the "canceled" state, so the mapping was never used. The supported "error" state would then not be recognized and a `ValueError` would be raised when attempting to map it. This has been fixed so that the "error" state is properly mapped.

RELEASE NOTES BEGIN

The `ogr` library now correctly handles the Forgejo 'error' commit flag, resolving an issue that previously caused it to fail during internal mapping.

RELEASE NOTES END

### DIFF
--- a/ogr/services/forgejo/flag.py
+++ b/ogr/services/forgejo/flag.py
@@ -24,7 +24,7 @@ class ForgejoCommitFlag(BaseCommitFlag):
         "pending": CommitStatus.pending,
         "success": CommitStatus.success,
         "failure": CommitStatus.failure,
-        "canceled": CommitStatus.error,
+        "error": CommitStatus.error,
         "warning": CommitStatus.warning,
     }
 


lbarcziova (approved):

ogr/services/forgejo/flag.py
```diff
@@ -24,7 +24,7 @@ class ForgejoCommitFlag(BaseCommitFlag):
         "pending": CommitStatus.pending,
         "success": CommitStatus.success,
         "failure": CommitStatus.failure,
-        "canceled": CommitStatus.error,
+        "error": CommitStatus.error,
```
mfocko, line 27: There’s also `skipped`, not sure if we have a matching state for that:

https://codeberg.org/forgejo/forgejo/src/commit/07ba27a02b8db81be78e336f8597a355c04adb18/modules/structs/commit_status.go#L10-L23

mfocko (approved):

ogr/services/forgejo/flag.py
betulependule: Oh, I see, that one was added 2 months ago. Our `CommitStatus` Enum doesn't contain `skipped`, but I can make a small follow-up PR to add it there (and add it to the `ForgejoCommitFlag` as well).

betulependule: According to Claude, the latest version of pyforgejo, which ogr uses, doesn't support the `skipped` status and it's not present in Forgejo's actual commit-status API, either, so if we support this status in ogr and someone tried to set a commit status to `skipped`, it would lead to an error.

mfocko: Don’t tell me it’s a dead code 🤦

mfocko: Well, it’s wrong then :D
https://codeberg.org/api/swagger#/repository/repoCreateStatus (you need to expand the `CommitStatusState`)

mfocko: It was added 2 months ago: https://codeberg.org/forgejo/forgejo/commit/8a1021e2a08917b9ce8f2c286c09baa4a7a2887f

surprisingly the docs got (somehow) adjusted… I’ve already opened a PR for something like that in the past :D when they scratched a state from GitHub and got hit with the compatibility issues (but then, after the revert, they forgot about the docs)

betulependule: https://codeberg.org/harabat/pyforgejo/src/branch/main/pyforgejo/repository/types/repo_list_statuses_request_state.py

The `skipped` status isn't in pyforgejo's upstream project, though. And Claude tried downloading the latest release of pyforgejo, checked it and only found the following statuses: :eyes: 

- pending
- success
- error
- failure
- warning

PR #996, opened 2026-07-23
Update service mapping URLs in preparation for the dist-git migration

### LINKED ISSUE #997: Update `SERVICE_MAPPING` to support Forgejo dist-git
`ogr`'s [`get_service_class_or_none`](https://github.com/packit/ogr/blob/675ada6a71b3b02a6389082d8feb44a73ddaab14/ogr/factory.py#L113) method, which is expected to return the appropriate Service class instance based on a given hostname, needs updating in order to support the migration of dist-git. Right now, `src.fedoraproject.org` is mapped to `PagureService`. However, once the migration of dist-git takes place, this hostname should instead correspond to `ForgejoService`.

I asked the team working on the migration and they said they plan to migrate all repos at once, so the migration shouldn't be gradual per-repo (as is the case with Fedora Forge). This is still worth verifying before the migration takes place in case there is a change of plans so that we can coordinate with their team to make sure Packit's transition to Forgejo dist-git support is smooth.

~~Whether Packit supports Pagure or Forgejo as dist-git relies on `SERVICE_MAPPING`, which is hardcoded, meaning we would need to release ogr with these changes at the same time that the switch of dist-git takes place, which would be difficult to coordinate and there would be a delay. It would be ideal to figure out a way we can make this transition smooth on Packit's side.~~

An automatic solution that would probe the dist-git backend that would determine whether to use `PagureService` or `ForgejoService` wouldn't be very clean and would introduce unnecessary behavioral changes and API calls, so instead, the goal is to make a clean update of the static service mapping and release it after the migration of dist-git.

Dist-git URLs are currently used as default values of some attributes in `PagureService`. These should be replaced / removed.

TODO:
- [x] Update `SERVICE_MAPPING` to map the appropriate hostnames to `ForgejoService`
- [x] Remove dist-git URLs from `PagureService`
- [ ] Update tests (some will start to fail after changing the mapping)

betulependule:
Service mapping needs to be updated after the migration so that `Config.load_authentication()` in [packit](https://github.com/packit/packit/blob/33bc381e684e4a9c649f26d557620fd361f21532/packit/config/config.py#L173) is able to authenticate against Forgejo as dist-git.

- [ ] Ensure smooth transition from Pagure to Forgejo
- [ ] Update tests
- [ ] Create a follow-up issue (to be done after the migration of dist-git): remove all code relevant to the determination of dist-git forge service class (it will no longer be needed and packit would be making unnecessary API calls if the code is not removed)

Fixes [#997](https://github.com/packit/ogr/issues/997)
Related to [#2681](https://github.com/packit/packit/issues/2681)

Note: Ignore the name of the branch that this PR was created from. This PR is unrelated to Fedora Forge. I would rename the branch, but don't want to risk breaking the PR / references to the branch.

Merge right before / during the migration of dist-git from Pagure to Forgejo. This mustn't be merged now. I don't think it would break anything if this were merged now, but it would lead to (currently) unnecessary API calls to see whether `PagureService` is supposed to be used when working with dist-git. I'll keep this PR a draft until then to make sure this doesn't get accidentally merged.

### DIFF
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -6,7 +6,7 @@ import logging
 from collections.abc import Iterable
 from typing import Optional
 
-from requests.exceptions import ConnectionError
+from requests.exceptions import ConnectionError, ReadTimeout
 
 from ogr.abstract import GitProject, GitService
 from ogr.constant import DGIT_URLS
@@ -124,7 +124,7 @@ def get_service_class_or_none(
     Get the matching service class from the URL.
     When attempting to get the matching service class for dist-git, probing
     is used to determine whether `PagureService` or `ForgejoService`
-    should be returned.
+    should be returned. This probing is set to timeout after 5 seconds.
 
     Args:
         url: URL of the project, e.g. `"https://github.com/packit/ogr"`.
@@ -137,7 +137,7 @@ def get_service_class_or_none(
         Matched class (subclass of `GitService`) or `None`.
 
     Raises:
-        OgrNetworkError, in case a ConnectionError error
+        OgrNetworkError, in case a ConnectionError or ReadTimeout error
             is encountered when attempting to probe Pagure dist-git.
     """
     mapping = {}
@@ -169,14 +169,14 @@ def get_service_class_or_none(
 
             try:
                 pagure_service = PagureService()
-                response = pagure_service.get_raw_request(url=request_url)
+                response = pagure_service.get_raw_request(url=request_url, timeout=5)
 
                 # if not found, then dist-git is no longer hosted on Pagure
                 if response.status_code != 404:
                     return PagureService
                 return ForgejoService
 
-            except ConnectionError as er:
+            except (ConnectionError, ReadTimeout) as er:
                 logger.error(er)
                 raise OgrNetworkError(f"Cannot connect to url: '{url}'.") from er
 
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -168,12 +168,11 @@ def get_service_class_or_none(
             from ogr.services.forgejo import ForgejoService
             from ogr.services.pagure import PagureService
 
-            cache = _DGIT_FORGE_CACHE.get(dgit_url)
-
-            now = time.monotonic()
-            if cache and ((now - cache[0]) < _DGIT_FORGE_CACHE_TTL):
-                return cache[1]
-
+            if cache := _DGIT_FORGE_CACHE.get(dgit_url):
+                timestamp, service_type = cache
+                now = time.monotonic()
+                if now - timestamp < _DGIT_FORGE_CACHE_TTL:
+                    return service_type
             # API call to the Pagure backend
             api_endpoint = "https://src.fedoraproject.org/api/0/version"
             api_endpoint_stg = "https://src.stg.fedoraproject.org/api/0/version"
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -168,9 +168,9 @@ def get_service_class_or_none(
             from ogr.services.forgejo import ForgejoService
             from ogr.services.pagure import PagureService
 
+            now = time.monotonic()
             if cache := _DGIT_FORGE_CACHE.get(dgit_url):
                 timestamp, service_type = cache
-                now = time.monotonic()
                 if now - timestamp < _DGIT_FORGE_CACHE_TTL:
                     return service_type
             # API call to the Pagure backend
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -40,7 +40,7 @@ def use_for_service(service: str, _func=None):
         pass
 
     @use_for_service("pagure.io")
-    @use_for_service("src.fedoraproject.org")
+    @use_for_service("git.centos.org")
     class PagureService(BaseGitService):
         pass
     ```
@@ -240,13 +240,13 @@ def get_instances_from_dict(instances: dict) -> set[GitService]:
     ```py
     get_instances_from_dict({
         "github.com": {"token": "abcd"},
-        "pagure": {
+        "forgejo": {
             "token": "abcd",
             "instance_url": "https://src.fedoraproject.org",
         },
     }) == {
         GithubService(token="abcd"),
-        PagureService(token="abcd", instance_url="https://src.fedoraproject.org")
+        ForgejoService(token="abcd", instance_url="https://src.fedoraproject.org")
     }
     ```
 
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -3,6 +3,7 @@
 
 import functools
 import logging
+import time
 from collections.abc import Iterable
 from typing import Optional
 
@@ -14,6 +15,10 @@ from ogr.exceptions import OgrException, OgrNetworkError
 from ogr.parsing import parse_git_repo
 
 _SERVICE_MAPPING: dict[str, type[GitService]] = {}
+
+# cache dist-git forge service class for two minutes
+_DGIT_FORGE_CACHE: dict[str, tuple[float, type[GitService]]] = {}
+_DGIT_FORGE_CACHE_TTL = 120
 
 logger = logging.getLogger(__name__)
 
@@ -124,7 +129,8 @@ def get_service_class_or_none(
     Get the matching service class from the URL.
     When attempting to get the matching service class for dist-git, probing
     is used to determine whether `PagureService` or `ForgejoService`
-    should be returned. This probing is set to timeout after 5 seconds.
+    should be returned. This information is cached for two minutes. The
+    probing request is set to timeout after 5 seconds.
 
     Args:
         url: URL of the project, e.g. `"https://github.com/packit/ogr"`.
@@ -162,6 +168,12 @@ def get_service_class_or_none(
             from ogr.services.forgejo import ForgejoService
             from ogr.services.pagure import PagureService
 
+            cache = _DGIT_FORGE_CACHE.get(dgit_url)
+
+            now = time.monotonic()
+            if cache and ((now - cache[0]) < _DGIT_FORGE_CACHE_TTL):
+                return cache[1]
+
             # API call to the Pagure backend
             api_endpoint = "https://src.fedoraproject.org/api/0/version"
             api_endpoint_stg = "https://src.stg.fedoraproject.org/api/0/version"
@@ -172,9 +184,12 @@ def get_service_class_or_none(
                 response = pagure_service.get_raw_request(url=request_url, timeout=5)
 
                 # if not found, then dist-git is no longer hosted on Pagure
-                if response.status_code != 404:
-                    return PagureService
-                return ForgejoService
+                dgit_service_kls = (
+                    PagureService if response.status_code != 404 else ForgejoService
+                )
+
+                _DGIT_FORGE_CACHE[dgit_url] = (now, dgit_service_kls)
+                return dgit_service_kls
 
             except (ConnectionError, ReadTimeout) as er:
                 logger.error(er)
--- a/ogr/factory.py
+++ b/ogr/factory.py
@@ -2,14 +2,20 @@
 # SPDX-License-Identifier: MIT
 
 import functools
+import logging
 from collections.abc import Iterable
 from typing import Optional
 
+from requests.exceptions import ConnectionError
+
 from ogr.abstract import GitProject, GitService
-from ogr.exceptions import OgrException
+from ogr.constant import DGIT_URLS
+from ogr.exceptions import OgrException, OgrNetworkError
 from ogr.parsing import parse_git_repo
 
 _SERVICE_MAPPING: dict[str, type[GitService]] = {}
+
+logger = logging.getLogger(__name__)
 
 
 def use_for_service(service: str, _func=None):
@@ -116,6 +122,9 @@ def get_service_class_or_none(
 ) -> Optional[type[GitService]]:
     """
     Get the matching service class from the URL.
+    When attempting to get the matching service class for dist-git, probing
+    is used to determine whether `PagureService` or `ForgejoService`
+    should be returned.
 
     Args:
         url: URL of the project, e.g. `"https://github.com/packit/ogr"`.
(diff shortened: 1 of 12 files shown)


lbarcziova:
> Merge after the migration of dist-git from Pagure to Forgejo. This mustn't be merged now, otherwise it will break authentication against Pagure. I'll keep this PR a draft until then to make sure this doesn't get accidentally merged.
 
 could we make it work with both at the same time? I assume that would ease up testing and transition period.

mfocko:
> could we make it work with both at the same time? I assume that would ease up testing and transition period.

I don’t think both should be a problem, though… I we cannot have both Forgejo and Pagure under one domain, and also I’m not sure how would the syncing on Fedora Infra work… afaik the changes flow only from prod to stage on Fedora Infra, not the other way around… yet… we would submit production builds and changes, so… this would be a mess

betulependule:
> > Merge after the migration of dist-git from Pagure to Forgejo. This mustn't be merged now, otherwise it will break authentication against Pagure. I'll keep this PR a draft until then to make sure this doesn't get accidentally merged.
> 
> could we make it work with both at the same time? I assume that would ease up testing and transition period.

I think it should be possible to support both? Packit uses the [get_service_class_or_none](https://github.com/packit/ogr/blob/8a9787200dc24411e2e0294178c6def0cbb3bbef/ogr/factory.py#L113) to get the appropriate service class, which would (in the current implementation) only return an instance of PagureService or ForgejoService (whichever it got to first in the for loop). This function could be modified so that ogr checks which one should be used, or we could use a new function, which would return all candidate service classes and packit would handle the checks.

If the migration of all repos happens at once, we can check the backend at runtime: `GET src.fedoraproject.org/api/0/version` (Pagure) and `GET src.fedoraproject.org/api/v1/version` (Forgejo) to see, which returns OK 200.

Or if the migration is gradual per-repo, we would need `GET src.fedoraproject.org/api/v1/rpms/<package>` to see whether the Forgejo backend is in use for the given repo at runtime. I'm not sure about the amount of  API calls this would lead to. We would possibly need caching.

betulependule:
I asked the team behind the migration and I was told they plan for the migration of all repos to happen all at once. They will migrate all repos to a new deployment and once that is done, make `src.fp.o` point to the new deployment.

It would still be problematic for us if we simply edit `SERVICE_MAPPING` during the migration, because we would then need to wait for the new release of `ogr` for Packit to use the updated mapping and actually work for Forgejo dist-git.

I suppose we could:
1. Right before the migration, release `ogr` with the following change: The [get_service_class_or_none()](https://github.com/packit/ogr/blob/8a9787200dc24411e2e0294178c6def0cbb3bbef/ogr/factory.py#L113) and [get_service_class()](https://github.com/packit/ogr/blob/8a9787200dc24411e2e0294178c6def0cbb3bbef/ogr/factory.py#L143) methods make API calls to check whether dist-git is currently hosted on Pagure or Forgejo, and return the appropriate class when needed (`PagureService` or `ForgejoService`). 
2. Once the migration is done, we release `ogr` once again, remove the API calls, and put in a static `SERVICE_MAPPING` of `src.fedoraproject.org` to `ForgejoService`.

I can't say I like this solution and it would be very clunky with a lot of unnecessary API calls, so I hope there could be a better solution.

betulependule:
If my understanding is correct, the requre tests are failing because of the API call to https://src.fedoraproject.org/api/0/version (a lacking recording of a response to this GET request). Multiple files in the `packit` repo (in `packit/tests_recording/test_data/test_status/`) need to have this response added in order for tests to pass in this PR.

tests/unit/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+from flexmock import flexmock
+
+from ogr import PagureService
+
+
+# mocks API calls to Pagure dist-git made to determine whether dist-git
+# is still hosted on Pagure and returns the status code expected after
+# the migration of dist-git to Forgejo
+@pytest.fixture(autouse=True)
+def setup_api_request_mock():
```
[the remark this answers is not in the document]
betulependule: Sure, but I don't think it matters.

ogr/services/forgejo/service.py
```diff
@@ -3,20 +3,24 @@
...
 from ogr.factory import use_for_service
+from ogr.parsing import parse_git_repo
 from ogr.services.base import BaseGitService
 from ogr.services.forgejo.project import ForgejoProject
 from ogr.services.forgejo.user import ForgejoUser
 
 
 @use_for_service("forgejo")
 @use_for_service("codeberg.org")
+@use_for_service("src.fedoraproject.org")
```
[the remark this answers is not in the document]
betulependule: I don't think it matters whether the decorators are added to `ForgejoService`. If dealing with dist-git URLs, probing is always used to determine whether `ForgejoService` or `PagureService` should be used. The decorators should have no effect until the probing is removed.

betulependule:
I suppose this PR is getting way too big already. I'll separate the small general Forgejo fixes (to bugs discovered in this PR when updating tests) to separate, smaller PRs.

ogr/factory.py
```diff
@@ -126,13 +135,51 @@ def get_service_class_or_none(
...
+
     if service_mapping_update:
         mapping.update(service_mapping_update)
+        non_overridden_dgit_urls = (
+            set(non_overridden_dgit_urls) - service_mapping_update.keys()
+        )
 
     parsed_url = parse_git_repo(url)
+
+    # [XXX] remove once the migration of dist-git is finished
```
mfocko, line 161: Just an FYI:
- you need to keep in mind that this will probably get also released, so even if we remove it after the migration is finished, it might live somewhere for some time…
- also, it would be ideal to have this in some release **before** the migration

ogr/factory.py
```diff
@@ -126,13 +135,51 @@ def get_service_class_or_none(
...
+        # if dealing with dist-git, we need to check whether we need to use
+        # `PagureService` or `ForgejoService`
+        if dgit_url in parsed_url.hostname:
+
+            from ogr.services.forgejo import ForgejoService
+            from ogr.services.pagure import PagureService
+
+            # API call to the Pagure backend
+            api_endpoint = "https://src.fedoraproject.org/api/0/version"
+            api_endpoint_stg = "https://src.stg.fedoraproject.org/api/0/version"
```
mfocko, line 178: I was going to say that it doesn’t work for pagure.io, so it’s some dist-git specific patch, but… now that I checked the forge itself, I cannot seem to get there altogether…

ogr/factory.py
```diff
@@ -126,13 +135,51 @@ def get_service_class_or_none(
...
+            request_url = api_endpoint_stg if ".stg." in dgit_url else api_endpoint
+
+            try:
+                pagure_service = PagureService()
+                response = pagure_service.get_raw_request(url=request_url)
+
+                # if not found, then dist-git is no longer hosted on Pagure
+                if response.status_code != 404:
```
mfocko: #sarcastic 50x could also mean it’s a Pagure…

ogr/factory.py
```diff
@@ -162,6 +168,12 @@ def get_service_class_or_none(
             from ogr.services.forgejo import ForgejoService
             from ogr.services.pagure import PagureService
 
+            cache = _DGIT_FORGE_CACHE.get(dgit_url)
+
+            now = time.monotonic()
+            if cache and ((now - cache[0]) < _DGIT_FORGE_CACHE_TTL):
+                return cache[1]
+
```
mfocko:
```suggestion
            if cache := _DGIT_FORGE_CACHE.get(dgit_url):
                timestamp, service_type = cache
                now = time.monotonic()
                if now - timestamp < _DGIT_FORGE_CACHE_TTL:
                    return service_type
```

I was going to say that you could probably just take a library for the cache, but… looks like we only have TTL cache in Packit / Service, not in ogr. I’m not sure if this “manoeuvre” is worth adding another dependency (this would propagate all the way to the Fedora too…)

ogr/services/pagure/service.py
```diff
@@ -32,7 +32,7 @@ class PagureService(BaseGitService):
     def __init__(
         self,
         token: Optional[str] = None,
-        instance_url: str = "https://src.fedoraproject.org",
+        instance_url: str = "https://pagure.io",
```
mfocko, line 35: this is going to die too tbf, not sure if before or after dist-git

tests/unit/test_pagure.py
```diff
@@ -8,5 +8,8 @@
 
 class TestPagureService(TestCase):
     def test_hostname(self):
-        assert PagureService().hostname == "src.fedoraproject.org"
-        assert PagureService(instance_url="https://pagure.io").hostname == "pagure.io"
+        assert PagureService().hostname == "pagure.io"
+        assert (
+            PagureService(instance_url="https://git.centos.org").hostname
```
mfocko, line 13: this seems to be down already… for the tests it doesn’t matter I think

ogr/services/pagure/service.py
betulependule: Yep, as you pointed out, "pagure.io" is read-only now. I could change `instance_url` to "git.centos.org", but if that's going to die as well, then I might as well keep it set to "src.fedoraproject.org".

ogr/factory.py
betulependule: I don't think it would be worth it to add another dependency. This code (hopefully) should be removed as soon as it's no longer needed.

Thanks for the code suggestion. It looks much more readable than my version. 🙏

ogr/factory.py
betulependule: Are you sure? I can get to "https://src.stg.fedoraproject.org/api/0/version" just fine.

tests/unit/test_pagure.py
betulependule: Perhaps I could just mark it with `@pytest.mark.skip`?

ogr/factory.py
```diff
@@ -126,13 +141,59 @@ def get_service_class_or_none(
...
+        if dgit_url in parsed_url.hostname:
+
+            from ogr.services.forgejo import ForgejoService
+            from ogr.services.pagure import PagureService
+
+            if cache := _DGIT_FORGE_CACHE.get(dgit_url):
+                timestamp, service_type = cache
+                now = time.monotonic()
+                if now - timestamp < _DGIT_FORGE_CACHE_TTL:
+                    return service_type
```
mfocko, line 175:
```suggestion
            now = time.monotonic()
            if cache := _DGIT_FORGE_CACHE.get(dgit_url):
                timestamp, service_type = cache
                if now - timestamp < _DGIT_FORGE_CACHE_TTL:
                    return service_type
```

I didn’t realise you’re using it below:
```
        # [XXX] remove once the migration of dist-git is finished
        for dgit_url in non_overridden_dgit_urls:
    
            # if dealing with dist-git, we need to check whether we need to use
            # `PagureService` or `ForgejoService`
            if dgit_url in parsed_url.hostname:
    
                from ogr.services.forgejo import ForgejoService
                from ogr.services.pagure import PagureService
    
                if cache := _DGIT_FORGE_CACHE.get(dgit_url):
                    timestamp, service_type = cache
                    now = time.monotonic()
                    if now - timestamp < _DGIT_FORGE_CACHE_TTL:
                        return service_type
                # API call to the Pagure backend
                api_endpoint = "https://src.fedoraproject.org/api/0/version"
                api_endpoint_stg = "https://src.stg.fedoraproject.org/api/0/version"
                request_url = api_endpoint_stg if ".stg." in dgit_url else api_endpoint
    
                try:
                    pagure_service = PagureService()
                    response = pagure_service.get_raw_request(url=request_url, timeout=5)
    
                    # if not found, then dist-git is no longer hosted on Pagure
                    dgit_service_kls = (
                        PagureService if response.status_code != 404 else ForgejoService
                    )
    
>                   _DGIT_FORGE_CACHE[dgit_url] = (now, dgit_service_kls)
                                                   ^^^
E                   UnboundLocalError: cannot access local variable 'now' where it is not associated with a value

/usr/lib/python3.15/site-packages/ogr/factory.py:190: UnboundLocalError
```

ogr/factory.py
mfocko: pagure.io, not staging dist-git

ogr/factory.py
betulependule: Thanks. 🙏 I didn't get around to fix it yesterday.

ogr/services/pagure/service.py
mfocko: (best would be _no default_, but… that is breaking change… idk, the pagure.io is probably the most sane default)

tests/unit/test_pagure.py
mfocko: I think you can leave it

ogr/services/pagure/service.py
betulependule: Well, changing `instance_url` from "https://src.fedoraproject.org" to anything else is a breaking change regardless. I'm thinking that once this is released before the migration, whoever is using `PagureService` without explicitly setting `instance_url` to "https://src.fedoraproject.org" will run into an error, despite dist-git still being on Pagure at that point. Maybe it's safer to keep the default value set to "https://src.fedoraproject.org" and only change it in a follow-up PR after the migration to make sure no users are affected  before the migration.

betulependule: As fullsend pointed out, another breaking change similar to this one is in `pagure/project.py‎` where removing the hardcoded dist-git URLs would break the `is_private` method whenever called pre-migration.

I think it would be best to drop `0506e3d7c9a45c16a2f9ef2fcce5097bd5b2bf1e` and `d044e7c6932f22179a4e2bf08ef599ac1f7d0276` from this PR, then create a follow-up PR with these two commits and merge it after migration, not before.

betulependule:
Closing as probing introduces a 5 second timeout on the additional the HTTP request to the affected methods, and can raise errors, changing the methods' behavior. It would also lead to unnecessary API calls, which would be introduced into at least one release of ogr.

Instead of probing, there will be a release of ogr (released after the migration of dist-git), which cleanly switches dist-git mapping from Pagure to Forgejo.

PR #1000, opened 2026-08-05
Fix the value of `_self.token` in `ForgejoService`

betulependule:
The value was previously set to f"token {token}", so when `token` was `None`, `self._token` would be set to "token None", which could lead to confusing behavior. `self._token` should be set to match `token` (to match the implementation of other forges). When instantiating `PyforgejoApi`, `api_key` needs to be set to an arbitrary string when making an unauthenticated request, otherwise a `ValueError` would be raised.

This issues was discovered when working on [#996](https://github.com/packit/ogr/pull/996).

RELEASE NOTES BEGIN

A bug has been fixed in relation to `ForgejoService` where `self._token` would be set to "token None" when it should be set to `None` instead.

RELEASE NOTES END

### DIFF
--- a/ogr/services/forgejo/service.py
+++ b/ogr/services/forgejo/service.py
@@ -28,12 +28,14 @@ class ForgejoService(BaseGitService):
     ):
         super().__init__()
         self.instance_url = instance_url + self.version
-        self._token = f"token {token}"
+        self._token = token
         self._api = None
 
     @cached_property
     def api(self) -> PyforgejoApi:
-        return PyforgejoApi(base_url=self.instance_url, api_key=self._token)
+        # pyforgejo doesn't add the "token " prefix by default
+        api_key = f"token {self._token}"
+        return PyforgejoApi(base_url=self.instance_url, api_key=api_key)
 
     def get_project(  # type: ignore[override]
         self,


ogr/services/forgejo/service.py
```diff
@@ -28,12 +28,14 @@ def __init__(
     ):
         super().__init__()
         self.instance_url = instance_url + self.version
-        self._token = f"token {token}"
+        self._token = token
```
mfocko, line 31: It doesn’t look like pyforgejo prefixes the token with the `token `, is this change intended?

mfocko: I’m kinda against it…

All implementations hide this implementation detail, e.g., the `forgejo-api` used in the Forgejo CLI:
https://codeberg.org/Cyborus/forgejo-api/src/commit/f9cf96b5fc76789b71b35b5924527c19f58d5c46/src/lib.rs#L368-L371

betulependule: When I first looked at the code, it seemed like setting `self._token` to f"token {token}" was inconsistent with the other forges, but now that I look at it again, it seems is done on purpose because all other forges already prefix the token with "token ". I'll go ahead and close this PR.

mfocko: I’m not against the change in handling _no provided token_, just this part of exposing the technical detail.

betulependule: Ok, I'll take a further look tomorrow.

betulependule: Or if we don'd want to set `self._token = f"token {token}"`, we could just set it to `self._token = token` and then add the "token " prefix in the `api` method instead of setting the `api_key` parameter to "unused".

PR #1011, opened 2026-08-26
Add support for allowing edits from maintainers on PRs/MRs

betulependule:
### PR summary
This PR adds support for enabling edits from maintainers on PRs/MRs across all forges, which support this (GitHub, GitLab and Forgejo). The Forgejo API doesn't allow setting this directly when creating a PR, so in this case, it is only supported when editing the PR created. Pagure doesn't support this option.

This functionality is needed in preparation for the migration of dist-git to Forgejo, but it was implemented for the other forges as well for consistency's sake.

While working on this, I also improved the implementation of `PullRequest.update_info()` for GitHub, which seemed problematic as it could raise an error. The fix is related to this code and is small enough that I don't think it would deserve a separate PR, so I included it as a separate commit.

### Tests
Integration tests have been added for GitHub and GitLab. 

Our Forgejo testing repo is hosted on `v10.next.forgejo.org`, which is past its EOL, so I was not able to create an account there to be able to run and record an integration test for Forgejo. I would generally use our packit-validator account for creating tests, but for this particular test, I would need a different account in order to create a fork and a PR from it.

As for Pagure, the tests are hosted on `pagure.io`, which is currently read-only, so I was not able to run a test for it, either. Considering that Pagure doesn't support enabling edits from maintainers anyway, this isn't a big deal.

Related to: [#2748](https://github.com/packit/packit/issues/2748)

RELEASE NOTES BEGIN

In `ogr`, a potential bug has been fixed in the GitHub implementation of `PullRequest.update_info()`. Previously, `None` values could be passed to the underlying method called to edit a PR, which could result in an `AssertionError` being raised as this method doesn't support `None` values.

`ogr` now supports the `allow_maintainer_edit` option for PRs/MRs for GitHub, GitLab and Forgejo. As the Forgejo API doesn't allow to set this directly when creating a PR, `allow_maintainer_edit` can only be set when updating it.

RELEASE NOTES END

### DIFF
--- a/tests/integration/gitlab/test_pull_requests.py
+++ b/tests/integration/gitlab/test_pull_requests.py
@@ -97,6 +97,30 @@ class PullRequests(GitlabTests):
         closed_pr = pr_for_closing.close()
         assert closed_pr.status == PRStatus.closed
 
+    def test_pr_maintainer_edits(self):
+        project = self.service.get_project(
+            repo="ogr-tests",
+            namespace=self.service.user.get_username(),
+        )
+
+        pr = project.create_pr(
+            title="test mainainer edits setting",
+            body="Description",
+            target_branch="master",
+            source_branch="pr-test1",
+            allow_maintainer_edit=True,
+        )
+
+        pr = self.project.get_pr(pr.id)
+
+        pr.update_info(description="this update shouldn't disable maintainer edits")
+        assert pr.allow_maintainer_edit
+
+        pr.update_info(allow_maintainer_edit=False)
+        assert not pr.allow_maintainer_edit
+
+        pr.close()
+
     def test_pr_merge(self):
         """
         Create new PR and update pull request ID to this test before this test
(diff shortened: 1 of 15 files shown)


tests/integration/gitlab/test_pull_requests.py
```diff
@@ -97,6 +97,30 @@ def test_pr_close(self):
...
+
+        pr = project.create_pr(
+            title="test mainainer edits setting",
+            body="Description",
+            target_branch="master",
+            source_branch="pr-test1",
+            allow_maintainer_edit=True,
+        )
+
+        pr = self.project.get_pr(pr.id)
```
lbarcziova, line 114: shouldnt this be called on `project.` instead of `self.project.` ?

lbarcziova (approved):
LGTM, just one note

tests/integration/gitlab/test_pull_requests.py
betulependule: `project` in this case refers to the fork project while `self.project` references the original repo. `pr.id` identifies the PR within the original repo, so using `project` wouldn't work.

PR #1016, opened 2026-09-14
ci(packit): run Koji/Bodhi on production

mfocko:
Move Koji/Bodhi automation to the production, so that we are able to test the Forgejo dist-git deployment before it gets officially moved.

TODO:
- [ ] Distribute to dist-git branches once merged (to make sure there are no inconsistencies)


betulependule (approved):
LGTM. Bodhi update isn't part of `ogr`'s `.packit.yaml` so maybe I would consider removing the mention of it from the title of the PR and the commit?

